PR #4, opened 2018-05-24
Multi edge trigger

ggggggggg:
This adds an experimental `EdgeMulti` triggering mode with tests. It's incompatible with other trigger modes, and only triggers on rising edges, but I think it's worth merging in so we can test it with actual data sources. It's works basically how we discussed
1. Look for a difference between two points larger than a threshold. Call this i_potential. In my testing offline I found that if you do the kink model, it doesn't matter much how you got i_potential, so I used the easiest method.
2. Look for a local maximum following i_potential.
3. Use  6 samples before and 3 samples after i_potential to fit a kink model looking at values -1,-0.5-,0.5,1 relative to i_potential. Note kbest as a trigger.
4. Go through all sets of t,u,v where t is the index of the last trigger, u is the current trigger, v is the next trigger, and make records. There are a couple of settings, but the MakeShortRecords setting makes records greedy after a trigger, and makes records as large as possible with no shared samples.

It would be pretty easy to support negative going edges, and it would fairly easy to support auto triggers of some sort coexisting.

`AnalyzeData` works with records of `length<=dsp.NSamples`. Note that the standard analysis quantities are almost all NaN when `rec.presamples=0`

### DIFF
--- a/process_data.go
+++ b/process_data.go
@@ -11,15 +11,16 @@ import (
 
 // DataStreamProcessor contains all the state needed to decimate, trigger, write, and publish data.
 type DataStreamProcessor struct {
-	Channum     int
-	Publisher   chan<- []*DataRecord
-	Broker      *TriggerBroker
-	NSamples    int
-	NPresamples int
-	SampleRate  float64
-	LastTrigger FrameIndex
-	stream      DataStream
-	projectors  mat.Dense
+	Channum              int
+	Publisher            chan<- []*DataRecord
+	Broker               *TriggerBroker
+	NSamples             int
+	NPresamples          int
+	SampleRate           float64
+	LastTrigger          FrameIndex
+	LastEdgeMultiTrigger FrameIndex
+	stream               DataStream
+	projectors           mat.Dense
 	// realtime analysis is disable if projectors .IsZero
 	// otherwise projectors must be size (nbases,NSamples)
 	// such that projectors*data (data as a column vector) = modelCoefs
@@ -79,24 +80,6 @@ type DecimateState struct {
 	DecimateLevel   int
 	Decimate        bool
 	DecimateAvgMode bool
-}
-
-// TriggerState contains all the state that controls trigger logic
-type TriggerState struct {
-	AutoTrigger bool
-	AutoDelay   time.Duration
-
-	LevelTrigger bool
-	LevelRising  bool
-	LevelLevel   RawType
-
-	EdgeTrigger bool
-	EdgeRising  bool
-	EdgeFalling bool
-	EdgeLevel   int32
-
-	// TODO:  Noise info.
-	// TODO: group source/rx info.
 }
 
 // ConfigurePulseLengths sets this stream's pulse length and # of presamples.
--- a/process_data.go
+++ b/process_data.go
@@ -165,20 +165,16 @@ func (dsp *DataStreamProcessor) AnalyzeData(records []*DataRecord) {
 	var modelCoefs mat.VecDense
 	var modelFull mat.VecDense
 	var residual mat.VecDense
-	var dataVec mat.VecDense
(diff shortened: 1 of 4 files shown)


process_data.go
```diff
@@ -206,22 +185,29 @@ func (dsp *DataStreamProcessor) AnalyzeData(records []*DataRecord) {
...
-			}
-
+			// to support variable length records, we have to truncate projectors and basis
+			truncFront := dsp.NPresamples - rec.presamples
+			truncBack := dsp.NSamples - len(rec.data) - truncFront
+			i := 0
+			k := nbases
+			j := truncFront
+			l := cols - truncBack
+			projectors := dsp.projectors.Slice(i, k, j, l)
```
joefowler: Problem: the truncation of a basis does not produce projectors that are truncations of the originals. Our choices are:

1. Don't (yet) allow variable-length records. I know they will be important someday, but I am afraid that learning/deciding how to handle them might bog down this project considerably.

1. Have Dastard store projectors for all allowed lengths. (This might mean that we designate a set of allowed lengths, so that 500, 520, 540 are all allowed, while a record that you might want to be of length 534 has to be truncated to 520.)

1. Perform whitening or approximate whitening on the data in dastard. If you do this, then I think projection is still not as simple as truncating the projector, but it might be somewhat simpler.

triggering_test.go
```diff
@@ -456,3 +583,21 @@ func BenchmarkLevelTrigger0TriggersOpsAreSamples(b *testing.B) {
...
+func TestKinkModel(t *testing.T) {
+	xdata := []float64{0, 1, 2, 3, 4, 5, 6, 7}
+	ydata := []float64{0, 0, 0, 0, 1, 2, 3, 4}
+	ymodel, a, b, c, X2, err := kinkModelResult(3, xdata, ydata)
+	if a != 0 || b != 0 || c != 1 || X2 != 0 || err != nil {
+		t.Errorf("a %v, b %v, c %v, X2 %v, err %v, ymodel %v", a, b, c, X2, err, ymodel)
+	}
+	ymodel, a, b, c, X2, err = kinkModelResult(4, xdata, ydata)
+	if a != 0.6818181818181821 || b != 0.22727272727272738 ||
+		c != 1.1363636363636362 || X2 != 0.45454545454545453 || err != nil {
```
joefowler: It seems sketchy to do tests on floating point equality. I don't see any go-supported test for approximate equality, though, not in the standard libraries. Not sure what to suggest.

triggering.go
```diff
@@ -4,24 +4,269 @@ import (
...
+
+func kinkModel(k float64, x float64, a float64, b float64, c float64) float64 {
+	if x < k {
+		return a + b*(x-k)
+	} else {
+		return a + c*(x-k)
+	}
+}
+
+// takes k, an index into xdata and into ydata
```
joefowler: Need a clearer comment, especially as to the meaning of k. I assume it's the real-number location of the kink, but I don't think the above line says that.

Also does't golint insist that comments be of the form?
```go
// kinkModelResult does x,y,z...
// additional lines...
```
Actually, that might only be comments on exported functions.

triggering_test.go
ggggggggg: I agree in principle, but this work, and is simple, and they don't provide anything to help. I guess we could write our own approx tester or find one in a package, but lets wait till its a problem.

process_data.go
ggggggggg: I thought you told me we could just truncate, but given that we can't, you're right so I took this out. I left in all the changes that make the summary quantities work with variable length records, and the tests on those quantities.

ggggggggg:
I pushed a commit that addressed your comments. Thanks for the feedback. I meant to split my commit into a few with different message, but I messed that part up.

PR #8, opened 2018-06-01
File writing

ggggggggg:
Implements LJH22, LJH3, PubRecords, PubSummaries as publishing methods. Each is accessed with a similar API. If `dp` is a `DataPublisher` you would start writing to LJH3 with `dp.SetLJH3(args)`, check to see if LJH3 is being written with `dp.HasLJH3()` and stop with `dp.CloseLJH3()`. Simply replace `LJH3` with `LJH22` to do the same for LJH22.  The `SetPubX()` methods don't take arguments.

If you read the comments you'll see I had many problems with getting tests for zmq working, so I simplified. The test send messages, I just don't set up a sub port to receive them. I did some benchmarking and found that `czmq.Channeler` is pretty slow, so I just launch a goroutine that reads records from a `chan` and publishes. This is what you had done before, and I think it is sufficient to follow the zmq thread safety rules (one socket should live on one thread).


ggggggggg:
The basic approach is that there is a DataPublisher type, and it has a field for LJH22 files and a field for each other type of output. If the field is zero valued (nil), no output of that type is generated. If the field has value, then output of that type is generated. Currently I have two fields, but plan to add two more for LJH3 and OFF files.

ggggggggg:
I'm having a hard time making tests relying on sending and recieving messages with czmq reliable. On my travel mac I have yet to get a test using tcp ports working. So I switched to using inproc for the tests, but but it I have to add some extraneous code that helps for unknown reasons, and it still fails on occasion. 

It's not clear how worrisome this is, overall I'd say the goczmq package appears to to have a well designed API, and great documentation and tests, so I find it very unlikely it actually barely works. So I wonder if there is something I'm missing (an init call? I guess then it's doesn't have great documentation), or if there is some finite setup time (seconds????) or something.

ggggggggg:
I ran the gozmq Benchmarks, and found

```
BenchmarkChanneler-4            	   30000	     37173 ns/op	  27.55 MB/s
```

Which says it takes 37 us to send a 1024 byte packet in their Benchmark. That's pretty bad. Maybe I should be worried.

publish_data.go
```diff
@@ -3,33 +3,74 @@ package dastard
...
+				// if the file doesn't exists yet, create it and write header
+				err := dp.LJH22.CreateFile()
+				if err != nil {
+					return err
+				}
+				dp.LJH22.WriteHeader()
+			}
+			nano := record.trigTime.UnixNano()
+			data := make([]uint16, len(record.data))
+			for i, v := range record.data {
```
joefowler: I think you want go's copy() builtin function here. It might not have a different effect but is more idiomatic.

publish_data.go
```diff
@@ -38,21 +39,54 @@ func (dp *DataPublisher) RemoveLJH22() {
...
+}
+
+func (dp *DataPublisher) SetPubFeederChan() {
+	dp.PubFeederChan = PubFeederChan
+}
+func (dp *DataPublisher) RemovePubFeederChan() {
+	dp.PubFeederChan = nil
+}
+
+var PubFeederChan chan []*DataRecord
```
joefowler: Why do you have this as a package-global variable, and then you sometimes use it within each DataPublisher and sometimes don't? Why not just have a DataPublisher attribute (or whatever go calls them) that's a bool to indicate whether to use it? 

Probably because multiple DataPublishers need to share this channel, but this somehow feels like the wrong way. I can't really put my finger on it.

ljh/ljh.go
```diff
@@ -174,6 +175,100 @@ func (w *Writer) WriteRecord(rowcount int64, timestamp int64, data []uint16) err
...
+		return errors.New("header already written")
+	}
+	h := Header{Frameperiod: w.Timebase, Format: "LJH3", FormatVersion: "3.0.0",
+		TDM: HeaderTDM{NumberOfRows: w.NumberOfRows, NumberOfColumns: w.NumberOfColumns,
+			Row: w.Row, Column: w.Column}}
+	s, err := json.MarshalIndent(h, "", "    ")
+	if err != nil {
+		panic("MarshallIndent error")
+	}
+	_, err = w.file.Write(s)
```
joefowler: I think it's more idiomatic here and in several following lines to save a line and say, e.g.,

```
if _,err = w.file.Write(s); err != nil {
    return err
}
```

This only works if you don't need to save either of the 2 return value from `w.file.Write`, because they would go out of scope once the if statement ends.

publish_data.go
```diff
@@ -96,8 +117,8 @@ func (dp DataPublisher) PublishData(records []*DataRecord) error {
 	if dp.HasPubFeederChan() {
 		dp.PubFeederChan <- records
 	}
-	for _, record := range records {
-		if dp.HasLJH22() {
+	if dp.HasLJH22() {
```
joefowler: Yes, definitely do a loop over records inside each branch of the form `if dp.HasWhatever() {...}` The new way is smarter.

publish_data.go
```diff
@@ -61,34 +61,59 @@ func (dp *DataPublisher) RemoveLJH22() {
 	dp.LJH22.Close()
 	dp.LJH22 = nil
 }
+
 func (dp *DataPublisher) HasPubRecords() bool {
 	return dp.PubRecords != nil
 }
 
+// SetPubRecords starts publishing records with czmq over tcp at port=PortTrigs
```
joefowler: Yeah, you really should start clearing out all the Lint warnings about lack of doc strings. It's not like it will be easier later, once you've forgotten what all the functions do.

publish_data.go
```diff
@@ -130,22 +155,55 @@ func (dp DataPublisher) PublishData(records []*DataRecord) error {
...
-	// 32 bits: # of samples, total
-	// 32 bits: sample period, in seconds (float)
-	// 32 bits: volts per arb conversion (float)
-	// 64 bits: trigger time, in ns since epoch 1970
-	// 64 bits: trigger frame #
+// messageSummaries makes a message with the following format for publishing on portTrigs
+// Structure of the message header is defined in BINARY_FORMATS.md
+// 16 bits: channel number
+//  8 bits: header version number
+//  8 bits: Presamples
```
joefowler: The number of bits listed in these comments, and the list of comments, is inconsistent with the `binary.Write()` calls below.

ggggggggg:
See https://github.com/ggggggggg/zmq_tests

Channeler is super slow, so lets just use the socket interface in ZMQ with a channel that sends batches of messages. It should work, I just won't test the receive part in Dastard.

## Timings on travel MacBookAir
  * `./wuserver` then `time ./wuclient` 1s, ~1e6 msg/s
  * `./wuserver_channeler` then `time ./wuclient` 220s, 4e5 msg/s
  * `./wuserver_czmq` then `time ./wuclient` 11s, ~1e6 msg/s
  * `./wuserver_czmq_batcher` then `time ./wuclient` 7s, ~1e6 msg/s 
On my travel laptop, running `./wuserver` then `time ./wuclient` takes about 10 seconds.

publish_data.go
ggggggggg: `record.data` is `[]RawType`, `data` is `[]uint16`, so `copy` doesnt work. I don't think `ljh` can import `RawType` because then there would be a circular `dastard`->`ljh`->`dastard` dependency. I think `RawType` may be a premature attempt to handle multiple datatypes, but it was outside the scope of what I was working on so I just made it work.

ggggggggg:
I added Benchmarks for publishing, and the performance is currently TERRIBLE. It takes about 100us to publish on all four outlets at once. I think I've tracked it down to `binary.Write`, which is used in all but `PubSummaries` I believe.

```
BenchmarkPublish/PubRecords-4         	   30000	     54647 ns/op	  36.60 MB/s
BenchmarkPublish/PubSummaries-4       	  200000	      5250 ns/op	 380.95 MB/s
BenchmarkPublish/PubLJH22-4           	   30000	     38453 ns/op	  52.01 MB/s
BenchmarkPublish/PubLJH3-4            	   30000	     54445 ns/op	  36.73 MB/s
BenchmarkPublish/PubAll-4             	   20000	    116157 ns/op	  17.22 MB/s
BenchmarkPublish/PubNone-4            	200000000	         9.53 ns/op	209953.09 MB/s
BenchmarkPublish/RawTypeToUint16-4    	 1000000	      1756 ns/op	1138.30 MB/s
BenchmarkPublish/binary.Write-4       	   30000	     41028 ns/op	  48.75 MB/s
```

This maybe makes sense given the disclaimer [here](https://golang.org/pkg/encoding/binary/):
```
This package favors simplicity over efficiency. Clients that require high-performance serialization, especially for large data structures, should look at more advanced solutions such as the encoding/gob package or protocol buffers.
```

ggggggggg:
I removed all uses of `binary.Write` in the hot loop of publishing data, this help a lot with the speed of the zmq publishing (though 5 us still seems too slow to me, it will be fast enough i guess). But LJH writing is still slow. I added benchmarks in ljh to make sure it's not some thing else in the dastard causing the problem, and it appears that writing LJH files is slow. I haven't figured out why. Current benchmarks 

```
BenchmarkPublish/PubRecords-4         	  500000	      3623 ns/op	 551.90 MB/s
BenchmarkPublish/PubSummaries-4       	  500000	      3891 ns/op	 513.99 MB/s
BenchmarkPublish/PubLJH22-4           	   50000	     27111 ns/op	  73.77 MB/s
BenchmarkPublish/PubLJH3-4            	   30000	     42279 ns/op	  47.30 MB/s
BenchmarkPublish/PubAll-4             	   20000	     85265 ns/op	  23.46 MB/s
BenchmarkPublish/PubNone-4            	200000000	         9.29 ns/op	215305.62 MB/s
BenchmarkPublish/RawTypeToUint16-4    	 1000000	      1451 ns/op	1377.83 MB/s
BenchmarkPublish/binary.Write-4       	   50000	     34656 ns/op	  57.71 MB/s
BenchmarkPublish/rawTypeToBytes-4     	500000000	         3.29 ns/op	1215137.78 MB/s
```

ggggggggg:
I replaced multiple `file.Write` calls to writing to a `bytes.Buffer` followed by a single `file.Write` call. This seems to help quite a bit. In julia I found that I could write to a file 1 byte at a time and get the same performance (fast!!) as writing many bytes at a time, that doesn't seem to be the case in go.

Dastard writes LJH files faster now, but still a lot slower than writing to ljh without dastard involved.

## dastard
```
BenchmarkPublish/PubLJH22-4           	   10000	     19012 ns/op	 105.19 MB/s
BenchmarkPublish/PubLJH3-4            	   10000	     17733 ns/op	 112.78 MB/s
```

## ljh (dastard not involved)
```
BenchmarkLJH22-4       	  100000	     10468 ns/op	 191.04 MB/s
BenchmarkLJH3-4        	  200000	     10673 ns/op	 187.38 MB/s
BenchmarkFileWrite-4   	  200000	      8726 ns/op	 229.18 MB/s
```

joefowler:
Hey, I don't know what to do with your changes in the last 24 hours. I'm sorry, but I didn't have time to read over them. Do you want them reviewed? Do you consider this ready to merge? What about #11? What order should we do them in?

ggggggggg:
Maybe I should use https://golang.org/pkg/bufio/ ?

ggggggggg:
Now I get fairly reasonable performance for writing files, though the benchmarks are a bit inconsistent. I wish I had read [this](https://hashrocket.com/blog/posts/go-performance-observations) earlier. The key performance issues I saw were:

  * `binary.Write` is slow
  * `file.Write` is unbuffered IO and slow for small writes, `bufio.Writer.Write` is buffered
  * `goczmq.Channeler` is slow

Once all of those are handled, the performance seem pretty reasonable.

Most recent timings on travel macbook air.

## dastard
```
BenchmarkPublish/PubRecords-4         	  300000	      4494 ns/op	 444.96 MB/s
BenchmarkPublish/PubSummaries-4       	  500000	      2673 ns/op	 748.03 MB/s
BenchmarkPublish/PubLJH22-4           	  200000	      8534 ns/op	 234.33 MB/s
BenchmarkPublish/PubLJH3-4            	  200000	     17616 ns/op	 113.53 MB/s
BenchmarkPublish/PubAll-4             	  100000	     17017 ns/op	 117.52 MB/s
```

## ljh
```
BenchmarkLJH22-4        	  300000	      8121 ns/op	 246.25 MB/s
BenchmarkLJH3-4         	  300000	      7978 ns/op	 250.67 MB/s
```

And if I follow the method [here](http://hints.macworld.com/article.php?story=20120704113548693) to benchmark the write speed on this laptop, I get 242 MB/s. So At least in some cases I'm writing as fast as this computer can to disk.

joefowler:
Thanks for your legwork on tracking down the buffering versus unbuffered methods of writing to file in Go. Although `bufio` is mentioned as early as page 9 in _The Go Programming Language_, the language isn't totally clear: "...the bufio package, which helps make input and output efficient and convenient." True, but it would have been nice to mention the buffering! Instead, it focuses on `bufio.Scanner`, which is good for parsing input into lines, words, whatever. All in all, that's a relief!

As you can see I have merged the branch viper into master and master into this one. It seems still to work, so if you are ready, I think we should merge this into master and close it.

PR #9, opened 2018-06-03
off format sketch

ggggggggg:
This is a quick sketch of what an off file writer could look like. It's almost just a copy and past of LJH3, which suggests they could share most code if we wanted. Though, it will be less work to copy and paste. 

Don't merge this, it needs more work (including tests). Mostly I just want to show you what it looks like so you can compare it to your `getdata` idea.


joefowler:
Looking at the 9-point list of ways we might write the projectors and the basis they project into, I think the ones that sound most appealing are the (1) MashallBinary (yes, it might be less automatic to read in other languages, but base64 is not exactly invented from scratch); (8) hdf5 files (but the "super con" sounds terrifying to me--if you, our local HDF5 expert, decides to make a go of it, I can't second-guess you, but it sounds big); or (9) dirfiles.

Regarding dirfiles and the fact that the system assumes a fixed data rate: I think we can work around this without a huge effort. We can generate a fake field to serve as the "reference field" (all other fields are assumed to be at least as long as that one). Just keep it as full as any of the arrays of pulse records. Other concerns are:

1. If every pulse is represented by 6 numbers, we have to decide whether to treat this as 6 separate fields or one field at 6x the rate. Or a hybrid: say, a field for the primary pulse-like projection, one for the background level, and a miscellaneous field to contain the others. I'll have to experiment with KST to see whether any of these are useful.
1. How to read the stored data in Julia. Will it require a full-blown Julia language binding? I experimented with this a few years ago but did only one or two basic function calls.

Possible advantages of dirfiles are:
- You can create synthetic channels that might represent Joule energy or something similar.
- You can then create a linear interpolation channel that implements a simple energy calibration.
- So KST or whatever is looking instantly at calibrated data, without us having to actually write energy estimates directly to a file.

Of course, you can do all this stuff with HDF5, too.

ggggggggg:
As I'm writing some code to load projectors and bases into dastard, I realize that there is a lot of useful into that isn't stored in DASTARD, like the noise model, projector covariance, the distributions of std_deviations_residuals of the training data, the training data filename.  All this sort make me lean towards 6. 6 has dastard commander writing the hdf5 file containing all this info to the off directory. Maybe it would be better to do that AND store the projectors and basis in the header of the off file?

I'm not sure if its a good idea to make sure you always have at least some info, or bad to allow you to make some progress with only some info, because it makes it easy for you to make the mistake of separating this info.

ggggggggg:
Closing to reduce noise while I work on this, will re-open.

joefowler:
Sounds good. I'm glad you're bringing this up to date. Also, you probably need a Vietnamese Noodle Bowl soon.

ggggggggg:
closing again because somehow the diff is super messy

PR #11, opened 2018-06-05
Use Viper for configuration management

joefowler:
I hope this doesn't conflict with #8, but it needs to be included. It's not final, but it's a good start at using Viper to manage saved configuration.

### DIFF
--- a/rpc_server.go
+++ b/rpc_server.go
@@ -167,16 +167,14 @@ func RunRPCServer(messageChan chan<- ClientUpdate, portrpc int) {
 	// Load stored settings
 	var okay bool
 	var spc SimPulseSourceConfig
-	fmt.Printf("viper used config file: %s\n", viper.ConfigFileUsed())
+	fmt.Printf("Dastard is using config file %s\n", viper.ConfigFileUsed())
 	err := viper.UnmarshalKey("simpulse", &spc)
 	if err == nil {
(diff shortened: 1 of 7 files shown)


joefowler:
p.s. If you do accept this PR overnight, feel free to close the branch viper.

rpc_server.go
```diff
@@ -134,6 +137,11 @@ func (s *SourceControl) broadcastUpdate() {
 	s.clientUpdates <- ClientUpdate{"STATUS", s.status}
 }
 
+// func (s *SourceControl) broadcastConfigurations() {
```
ggggggggg: this could be removed?

rpc_server.go
```diff
@@ -142,32 +150,46 @@ func (s *SourceControl) broadcastTriggerState() {
...
-	sourcecontrol := new(SourceControl)
-	sourcecontrol.clientUpdates = messageChan
-	server.Register(sourcecontrol)
+	sourceControl := new(SourceControl)
+	sourceControl.clientUpdates = messageChan
+
+	// Load stored settings
+	var okay bool
+	var spc SimPulseSourceConfig
+	fmt.Printf("Dastard is using config file %s\n", viper.ConfigFileUsed())
```
ggggggggg: this should use log

PR #17, opened 2018-06-15
Configurebasis

ggggggggg:
Add `SourceControl.ConfigureProjectorsBasis` to provide a way to set Projectors


rpc_server.go
```diff
@@ -49,7 +52,6 @@ func (s *SourceControl) Multiply(args *FactorArgs, reply *int) error {
 func (s *SourceControl) ConfigureTriangleSource(args *TriangleSourceConfig, reply *bool) error {
 	log.Printf("ConfigureTriangleSource: %d chan, rate=%.3f\n", args.Nchan, args.SampleRate)
 	err := s.triangle.Configure(args)
-	s.clientUpdates <- ClientUpdate{"TRIANGLE", args}
```
joefowler: Why this deletion and the next one (line 62)? This seems wrong.

rpc_server.go
```diff
@@ -119,17 +157,18 @@ func (s *SourceControl) Start(sourceName *string, reply *bool) error {
...
-		s.status.Running = false
-		s.status.SourceName = ""
-		s.status.Nchannels = 0
-	}
+	s.status.Running = false
+	s.status.SourceName = ""
+	s.status.Nchannels = 0
 	s.broadcastUpdate()
+	*reply = true
 	return nil
```
joefowler: This change make sense.

rpc_server.go
```diff
@@ -145,10 +184,13 @@ func (s *SourceControl) broadcastTriggerState() {
...
+// SendAllStatus causes a broadcast to clients containing all broadcastable status info
 func (s *SourceControl) SendAllStatus(dummy *string, reply *bool) error {
+	log.Println("A Client has requested to send all status")
+	s.broadcastTriggerState()
 	s.broadcastUpdate()
-	s.clientUpdates <- ClientUpdate{"SENDALL", 0}
+	s.broadcastTriggerState()
+	*reply = true
 	return nil
 }
```
joefowler: It looks to me like you are reverting to an earlier version of this method. I'm okay with your comment, but I think the function body should read 
```go
func (s *SourceControl) SendAllStatus(dummy *string, reply *bool) error {
	s.broadcastUpdate()
	s.clientUpdates <- ClientUpdate{"SENDALL", 0}
	return nil
}
```

PR #18, opened 2018-06-15
Fix auto trigger problem

joefowler:
Numbers segments correctly, and gives correct real-world read time.

This was a bear to figure out! Worse that solving the problem was creating a test that tested for it correctly.

### DIFF
--- a/simulated_data_sources.go
+++ b/simulated_data_sources.go
@@ -9,12 +9,11 @@ import (
 
 // TriangleSource is a DataSource that synthesizes triangle waves.
 type TriangleSource struct {
-	minval       RawType
-	maxval       RawType
-	timeperbuf   time.Duration
-	onecycle     []RawType
-	cycleLen     int
-	lastFrameNum FrameIndex
+	minval     RawType
+	maxval     RawType
+	timeperbuf time.Duration
+	onecycle   []RawType
+	cycleLen   int
 	AnySource
 }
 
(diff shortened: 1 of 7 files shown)


simulated_data_sources.go
```diff
@@ -84,9 +81,15 @@ func (ts *TriangleSource) blockingRead() error {
 	for _, ch := range ts.output {
 		datacopy := make([]RawType, ts.cycleLen)
 		copy(datacopy, ts.onecycle)
-		seg := DataSegment{rawData: datacopy, framesPerSample: 1}
+		seg := DataSegment{
+			rawData:         datacopy,
+			framesPerSample: 1,
+			firstFramenum:   ts.lastFrameNum,
+			firstTime:       ts.lastread,
+		}
```
ggggggggg: This looks like a case where a `NewDataSegment` function that requires all argument would have helped.

simulated_data_sources.go
```diff
@@ -183,9 +181,14 @@ func (sps *SimPulseSource) blockingRead() error {
 		for i := 0; i < sps.cycleLen; i++ {
 			datacopy[i] += RawType(rand.Intn(21) - 10)
 		}
-		seg := DataSegment{rawData: datacopy, framesPerSample: 1}
+		seg := DataSegment{
+			rawData:         datacopy,
+			framesPerSample: 1,
+			firstFramenum:   sps.lastFrameNum,
```
ggggggggg: This reads wrong? Shouldn't it be something like `lastFrameNum+1`? I'm guessing the answer is not, but a comment here explaining it would be nice.

joefowler: Can do. I'll think about a comment and whether the variable name is accurate.

simulated_data_sources.go
joefowler: Oddly enough, there is such a thing. It would have helped here, but I think I avoided it because it returned a `*DataSegment` instead of a `DataSegment`. Should I be passing around the latter instead of the former? Not sure.

PR #21, opened 2018-06-19
Mix

ggggggggg:
first pass at adding mix
I added a default implementation of `ConfigureMixFraction` to `AnySource` that errors, so that it can be part of the `interface` 

I'm not sure how to get the mix published on the updates channel, or maybe it already will be?

If you merge the export less internal state PR, this one should be simpler since it branches from that one

### DIFF
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -33,7 +33,8 @@ type LanceroSource struct {
 	ncards            int
 	clockMhz          int
 	active            []*LanceroDevice
-	chan2readoutOrder map[int]int
+	mixFraction       []float64
+	chan2readoutOrder []int // this could be a slice
 	AnySource
 }
 
@@ -41,7 +42,6 @@ type LanceroSource struct {
 func NewLanceroSource() (*LanceroSource, error) {
 	source := new(LanceroSource)
 	source.devices = make(map[int]*LanceroDevice)
-	source.chan2readoutOrder = make(map[int]int)
 	devnums, err := lancero.EnumerateLanceroDevices()
 	if err != nil {
 		return source, err
@@ -107,7 +107,11 @@ func (ls *LanceroSource) Configure(config *LanceroSourceConfig) error {
 // updateChanOrderMap updates the map chan2readoutOrder based on the number
 // of columns and rows in each active device
 func (ls *LanceroSource) updateChanOrderMap() {
-	ls.chan2readoutOrder = make(map[int]int)
+	allNChan := int(0)
+	for _, dev := range ls.active {
+		allNChan += dev.ncols * dev.nrows * 2
+	}
+	ls.chan2readoutOrder = make([]int, allNChan)
 	nchanPrevDevices := 0
 	for _, dev := range ls.active {
 		nchan := dev.ncols * dev.nrows * 2
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -411,6 +411,22 @@ func (ls *LanceroSource) distributeData(timestamp time.Time, wait time.Duration)
 					data[j] &= mask
 				}
 			} else { // apply mix
+				// comments based upon NDFBServer.cp
+				// fb_physical[n] refers to the feedback signal applied during tick [n]
+				// err_physical[n] refers to the error signal measured during tick [n], eg with fb_physical[n] applied
+				// fb_data[n]=fb_physical[n+1]
+				// err_data[n]=err_physical[n]
+				// in words: at frame [n] we get data for the error measured at frame [n]
+				// and the feedback that will be applied during frame [n+1]
+				// we want
+				// mix[n] = fb_physical[n] + mixFraction * err_physical[n]
+				// so
+				// mix[n] = fb_data[n-1]   + mixFraction * err_data[n]
+				// or
+				// mix[n+1] = fb_data[n]   + mixFraction * err_data[n+1]
+
+				// correct phasing not yet implemented
+
 				associatedErr := datacopies[ls.chan2readoutOrder[channum-1]]
 				for j := 0; j < len(data); j++ {
 					d := data[j] & mask
@@ -418,7 +434,6 @@ func (ls *LanceroSource) distributeData(timestamp time.Time, wait time.Duration)
 					data[j] = d + RawType(m)
 				}
 			}
-			// TODO: I think err->FB mixing goes here??
 		}
 
 		// TODO: replace framesPerSample=1 with the actual decimation level
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -33,8 +33,8 @@ type LanceroSource struct {
 	ncards            int
 	clockMhz          int
 	active            []*LanceroDevice
-	mixFraction       []float64
-	chan2readoutOrder []int // this could be a slice
+	chan2readoutOrder []int
+	MixFraction       []float64
 	AnySource
 }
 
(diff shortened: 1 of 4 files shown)


lancero_source.go
```diff
@@ -33,15 +33,15 @@ type LanceroSource struct {
 	ncards            int
 	clockMhz          int
 	active            []*LanceroDevice
-	chan2readoutOrder map[int]int
+	mixFraction       []float64
+	chan2readoutOrder []int // this could be a slice
```
joefowler: I made this a map instead of a slice on the theory that channel numbers might not be a contiguous set. I admit that this doesn't totally make sense. Also, I had the map backwards at first (that is, I mapped readout order to channel order, when it turned out we wanted the inverse map).

In other words, my thinking was muddled enough that I'll accept your idea to make this a slice instead of a map.

lancero_source.go
```diff
@@ -393,8 +405,18 @@ func (ls *LanceroSource) distributeData(timestamp time.Time, wait time.Duration)
 		// mask out frame and extern trigger bits from FB channels
 		if channum%2 == 1 {
 			const mask = ^RawType(0x3)
-			for j := 0; j < len(data); j++ {
-				data[j] &= mask
+			mixFraction := ls.MixFraction[channum]
```
joefowler: One point about mixing that you didn't know: the `ndfb_server` doesn't mix the current error with the current FB value. It mixes the _previous_ error into the current FB. (Or do I have that backwards??) I think the argument for doing this is both that it's more correct in some sense and that it is needed to match what we have been doing in past DAQ code.

Unfortunately, this involves special handling for the first sample, and saving one last error sample for each channel from call to call. Ugh!

lancero_source.go
```diff
@@ -125,6 +127,16 @@ func (ls *LanceroSource) updateChanOrderMap() {
 	}
 }
 
+// ConfigureMixFraction provides a default implementation for all non-lancero sources that
+// don't need the mix
+func (ls *LanceroSource) ConfigureMixFraction(processorIndex int, mixFraction float64) error {
+	if processorIndex >= len(ls.MixFraction) || processorIndex < 0 {
```
joefowler: I understand the simplicity of forcing clients to call this once per TES, but I'd like to suggest supporting the mode that if `processorIndex < 0` you set all mix values to the same level (presumably you'd only use this mode to set all mix to 0, but whatever).

lancero_source.go
ggggggggg: Yes i agree it is more annoying, and you're probably right that it is more correct. I'll double check with Carl, and think about how to handle it.

ggggggggg: Also I just went for it with the default plus override, I really hoped it would work, and though it might work, and was pleasantly surprised.

joefowler:
I just ran into Carl, and he verified that the meaning of the 4-byte package from the TDM system contains the error signal computed during a particular measurement and the FB signal _applied at the same time_. Thus the FB signal in one frame is the result of the error signal seen in all previous frames but is not affected by the current frame's error.

This doesn't explain what "the mix" is for, but it helps clarify the timing. He volunteered that one could imagine that we'd be better off if he had delayed the error signal by one sample, but he made a guess and we've gone with it for a decade.

joefowler:
Okay, now I just got Carl to say _the opposite thing_. Now he thinks that the error signal is bundled with the FB signal that is computed from it. That would make more sense, and better explain the off-by-1 alignment that the `ndfb_server` deals with.

Here is the relevant code from `NDFBServer.cpp`:

```c++
                // MIX: mix some fraction of the error value into the _previous_ FB value.

                // NB: at tick [n] we get the error at [n] and the fb for what it _will be_ during [n+1]
                // to achieve proper phasing, we will retard the fb stream by one tick s.t.  at time [n]
                // we get the fb that was applied at [n-1] and the error that was generated at [n] this
                // is important when mixing is on b/c otherwise you end up mixing the wrong error into
                // the feedback

                // Put another way, at tick [n] we get FB_in[n] and ER_in[n], but we don't want to mix
                // those to get the right phasing, we want

                // FB_out[n] = FB_in[n]   and ER[n] = ER[n+1]
                // problem is, we don't have ER[n+1] at tick [n]!, so retard the FB term 1 tick
                // FB_out[n] = FB_in[n-1] and ER[n] = ER[n]

                // mix the error and signal if necessary.  always _ADD_ the mix, if you want to switch
                // polarity, do it with the mixing inversion parameter


                if (fbStream[row]->mixFlag())
                {
                    float mixLevel = fbStream[row]->mixLevel();
                    if (fbStream[row]->mixInversionFlag())
                        mixLevel *= -1;
                    mixLevel /= coAddedSamples();

                    // Since "mix" mixes error[k] into the 1-sample-older feedback, we have
                    // shifted the rawFBSignals[] vector so that rawErrorSignals[k] and
                    // rawFBSignals[k] are the pairs that are to be mixed together.
                    for (int j=0; j<nvalues; j++)
                    {
                        // CAREFUL!  Use the round() function, else the mix will truncate towards zero
                        // That makes zero an "attractor" for all (mixLevel*error) between -0.999 and 0.999.
                        // That is twice the range of any other mixed-in integer.  Boo!
                        double mixed =  rawFBSignals[j] + round(mixLevel * (int16_t) rawErrorSignals[j]);

                        // Make sure that mix doesn't push FB out of the range 0-65535 (=UINT16_MAX).
                        if (mixed<0)
                            rawFBSignals[j] = 0;
                        else if (mixed > UINT16_MAX)
                            rawFBSignals[j] = UINT16_MAX;
                        else
                            rawFBSignals[j] = mixed;
                    }
                }
```

ggggggggg:
I think we're going to want a definitive from carl, I'm worried he's still wrong. Any way to confirm. Anyway, I think this plot tells us what we want to know: I think it says we should mix the error with the feedback value that was calculated using that error
figure_1-1

joefowler:
Galen: Randy and I disagreed with your words: "we should mix the error with the feedback value that was calculated using that error."  I'm not sure how your plot justifies them.

Randy's competing statement was: "we should mix the error with the feedback value that was applied at the moment that error was measured." Why? Because that is the error that shows how wrong our now-stale FB computation (really, a prediction) turned out to be.

ggggggggg:
"we should mix the error with the feedback value that was applied at the moment that error was measured." is what my plot actually shows, so easy to make a mistake. I agree it also makes intuitive sense.

joefowler:
Well, I'm glad you said it wrong, because honestly, I said it wrong at first. I initially typed "Randy and I agree 100% with your words." Then upon further reflection, I realized that I disagreed 100%. Very very confusing. This, incidentally, is another argument not to change what the NDFB server does!

ggggggggg:
This feels so hacky. I guess it would be more elegant to get Carl to change the firmware, but probably lead to problems with version control. It might be more elegant to deal with the data phasing logic at data distribution time?

Anyway, I updated it to account for the phasing.

joefowler:
It is pretty hacky. I should admit that Carl absolutely offered to change the firmware, and he was kind of apologetic about the way he arbitrarily decided to package the err+FB data a decade ago.

What do you mean "It might be more elegant to deal with the data phasing logic at data distribution time?" Isn't that what we're doing? 

Randy and Carl and I discussed the ramifications of changing firmware for a bit. I don't want to rule it out absolutely, but we are not in a great position to keep the revision control problems in order. The cards don't report their versioning in any accessible way. We could insist on John and Carl blasting new firmware to all lab systems, but then we'd also have to spread a new version of NDFB server around at the same time. And what about non-Boulder-based systems?

I think that to solve the alignment problem in dastard is hack-ish, but not so bad. It's straightforward to explain in comments. (Yeah, it's hard to get right, but it's simple enough to explain.) And it will be a non-issue once TDM systems are replaced by microwave MUX.

Are you ready to have me merge?

ggggggggg:
I'm ready to merge except in the sense that this hasn't actually been tested on hardware. It's probably better to wait until we can make sure it doesn't crash on hardware and the output looks roughly right.

By dealing with it during data distribution, I meant when making the slice called `datacopies`.

PR #22, opened 2018-06-19
Allow port numbers to vary, so testing can use different ports

joefowler:
This prevents tests from messing with a running dastard server.

### DIFF
--- a/networking.go
+++ b/networking.go
@@ -9,12 +9,12 @@ type Portnumbers struct {
 	Summaries      int
 }
 
-var ports = Portnumbers{5500, 5501, 5502, 5503, 5504}
+var Ports = Portnumbers{5500, 5501, 5502, 5503, 5504}
 
 func setPortnumbers(base int) {
-	ports.RPC = base
-	ports.Status = base + 1
-	ports.Trigs = base + 2
-	ports.SecondaryTrigs = base + 3
-	ports.Summaries = base + 4
+	Ports.RPC = base
+	Ports.Status = base + 1
+	Ports.Trigs = base + 2
+	Ports.SecondaryTrigs = base + 3
+	Ports.Summaries = base + 4
 }
--- a/networking.go
+++ b/networking.go
@@ -1,6 +1,6 @@
 package dastard
 
-// Portnumbers holds all TCP port numbers used by Dastard.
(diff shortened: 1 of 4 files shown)


networking.go
```diff
@@ -1,10 +1,20 @@
...
-// TCP port numbers used by Dastard.
-const (
-	PortRPC int = 5500 + iota
-	PortStatus
-	PortTrigs
-	PortSecondaryTrigs
-	PortSummaries
-)
+// Portnumbers holds all TCP port numbers used by Dastard.
+type Portnumbers struct {
```
ggggggggg: PortNumbers is more go styles i think

networking.go
```diff
@@ -1,10 +1,20 @@
...
+	RPC            int
+	Status         int
+	Trigs          int
+	SecondaryTrigs int
+	Summaries      int
+}
+
+var Ports = Portnumbers{5500, 5501, 5502, 5503, 5504}
```
ggggggggg: This should be `setPortnumbers(5500)`? Can you call a function in global scope? I guess it would need to be in an `init` function, then it would run after any `TestMain` and default your goal, so I guess it should stay as you wrote it.

joefowler: You can't call `setPortnumbers(5500)` at the package level. It could go in every main program, but that would be easy to forget.

What you _can_ apparently do is define a `function init() {...}` in each Go file, which will executed when the program starts. Such a function cannot be called other than automatically. Cool, huh? I think I'll use this idea and update the PR.

ggggggggg: What out because I think `init` runs after `TestMain` which means it may be hard to override ports in tests reliably.

joefowler:
The point is not that we _need_ to run tests while dastard is running, but that your editor (ahem, Atom) will silently run tests whenever you save a file, whether dastard is running or not. This will end up with the test program failing to start a second RPC server, and instead the test clients talk to your active server and completely reconfigure it. 

Right now this is merely annoying: your configuration gets overwritten by the settings used in the tests instead of the "real" settings. But if you were trying to take data and edit the code on the same machine, it could eventually cause loss of data and an aborted experiment. We don't want that.

Thanks for reviewing. I'll merge once the Travis CI runs finish. (Why is there a .../pr test and a .../push test?)

PR #24, opened 2018-06-20
Trigger control, Named channels, other changes

joefowler:
1. Allow trigger control over RPC from clients. 
1. Distribute a list of channel names for clients to use internally.
1. Reduced risk of race condition by better use of Mutex locks.
1. Change in Lancero data sampling policy: discard first buffer sample for better estimates of LSYNC.
1. ZMQ port numbers are not compile-time constants but variables. Allows us to run tests without messing up a running dastard server by feeding test commands into the main server.

Trigger control in Dastard-commander is not complete yet--we can control auto and level triggers so far. But this can be completed separate from the main dastard program.


data_source.go
```diff
@@ -114,6 +117,17 @@ func (ds *AnySource) Running() bool {
...
+	if len(ds.chanNames) < ds.nchan {
+		ds.chanNames = make([]string, ds.nchan)
+	}
+	for i := 0; i < ds.nchan; i++ {
+		if len(ds.chanNames[i]) == 0 {
+			ds.chanNames[i] = fmt.Sprintf("unk%d", i)
+		}
+	}
+}
+
```
ggggggggg: is this a defensive pre-fill? if so, a comment would be nice

ggggggggg: Or more descriptive name like defensivePreFillChannelNamesAsUnknown.

data_source.go
```diff
@@ -138,10 +153,11 @@ func (ds *AnySource) PrepareRun() error {
...
 	ds.processors = make([]*DataStreamProcessor, ds.nchan)
 	ds.runDone.Add(ds.nchan)
-	for chnum, ch := range ds.output {
-		dsp := NewDataStreamProcessor(chnum, ds.broker)
+	for idnum, dataSegChan := range ds.output {
+		dsp := NewDataStreamProcessor(idnum, ds.broker)
+		dsp.Name = ds.chanNames[idnum]
 		dsp.SampleRate = ds.sampleRate
-		ds.processors[chnum] = dsp
+		ds.processors[idnum] = dsp
```
ggggggggg: ive been calling this processorIndex, I think we should pick one name and use it consistently, im happy to have you make the decision without further bike shedding

ggggggggg: actually to match `channelNames` maybe we should use `channelIndex`?

PR #26, opened 2018-06-22
Lancero no hardware

ggggggggg:
* `lancero.NoHardware`
* refactor Mix
* above enables test coverage of 93.3%
* adds `od -xt` like output by default on every attempt to find frame bits

### DIFF
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -153,11 +153,8 @@ func (ls *LanceroSource) updateChanOrderMap() {
 // ConfigureMixFraction sets the MixFraction for the channel associated with ProcessorIndex
 // mix = fb + mixFraction*err
 func (ls *LanceroSource) ConfigureMixFraction(processorIndex int, mixFraction float64) error {
-<<<<<<< HEAD
-=======
 	ls.runMutex.Lock()
 	defer ls.runMutex.Unlock()
->>>>>>> master
 	if processorIndex >= len(ls.Mix) || processorIndex < 0 {
 		return fmt.Errorf("processorIndex %v out of bounds", processorIndex)
 	}
@@ -226,12 +223,7 @@ func (device *LanceroDevice) sampleCard() error {
 	const tooManyBytes int = 1000000  // shouldn't need this many bytes to SampleData
 	const tooManyIterations int = 100 // nor this many reads of the lancero
 	for i := 0; i < tooManyIterations; i++ {
-<<<<<<< HEAD
-		fmt.Println(i)
-		if bytesRead >= tooManyBytes && i > 1 { // we ignore first buffer, make sure we see 2nd
-=======
 		if bytesRead >= tooManyBytes && i > 1 { // we ignore first buffer, so make sure we've seen 2
->>>>>>> master
 			return fmt.Errorf("LanceroDevice.sampleCard read %d bytes, failed to find nrow*ncol",
 				bytesRead)
 		}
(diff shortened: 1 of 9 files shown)


joefowler:
Umm, these test failures aren't the usual failure to install a dependency. Did you check them?

```text
fatal error: runtime: out of memory
```

ggggggggg:
No I assumed they were the same, it passes on my computer. I guess it uses too much memory? I'll look into it this weekend. Should be fairly easy to limit memory usage, though I'm surprised it's very large.

joefowler:
Well, I had Travis catch an out-of-memory error a week or two ago. It was some kind of infinite loop, I think. That is, it would have run out of memory on any system. I think I missed it because I was running tests with a timeout on my machine. (Still, they must have _failed_, so it seems like I must have pushed when I should not have.)

lancero_source.go
```diff
@@ -259,6 +259,7 @@ func (ls *LanceroSource) StartRun() error {
 
 	// Starting the source for all active cards has 3 steps per card.
 	for _, device := range ls.active {
+		fmt.Println("loop start")
```
joefowler: You'll want to get rid of this and the next several `fmt.Print`-family lines.

lancero_source.go
```diff
@@ -37,7 +37,7 @@ type LanceroSource struct {
 	active            []*LanceroDevice
 	chan2readoutOrder []int
 	lastFbData        []RawType // used in mix calculation
-	MixFraction       []float64
+	Mix               []Mix
```
joefowler: I like the idea of the slice of objects that can recall both the amount of mix AND the necessary previous value.

mix.go
```diff
@@ -0,0 +1,50 @@
...
+// or
+// mix[n+1] = fb_data[n]   + mixFraction * err_data[n+1]
+type Mix struct {
+	mixFraction float64
+	lastFb      RawType
+}
+
+// MixRetardFb mixes err into fbs, alters fbs in place to contain the mixed values
+// consecutive calls must be on consecutive data
+func (m *Mix) MixRetardFb(fbs *[]RawType, errs *[]RawType) {
```
joefowler: I understand that you're not doing this any differently from how I did it, but do we need this copy? I have a sense that we could skip it, that it simplifying the loop but at a cost. Thoughts?

mix.go
```diff
@@ -0,0 +1,50 @@
...
+	unmixed := make([]RawType, len(*fbs))
+	unmixed[0] = m.lastFb
+	copy(unmixed[1:], (*fbs)[0:len(unmixed)-1])
+	m.lastFb = (*fbs)[len(unmixed)-1]
+	const mask = ^RawType(0x03)
+	for j := 0; j < len(*fbs); j++ {
+		fb := unmixed[j] & mask
```
joefowler: Wait, you already masked out the frame bit and the external trigger bit, on line 439 of `lancero_source.go`. Don't do it here, too!

mix.go
```diff
@@ -0,0 +1,50 @@
...
+		mixAmount := float64((*errs)[j]) * m.mixFraction
```
joefowler: I am uncomfortable keeping in this and all the following lines even when `m.mixFraction==0`. Maybe I'm micro-optimizing, but...

joefowler:
I think the problem is in `func TestNoHardware` in `lancero_test.go`. Your `defer source.Stop()` is not completing because of the wait group, yet somehow the data production bit is continuing to produce data. ?? Leaving soon, but I can try to help on Monday.

mix.go
ggggggggg: I tried to get rid of but failed a few times, I'm not convinced it's impossible though. I stopped working on it because the code is easier to read this way, and it didn't seem like a high priority.

mix.go
ggggggggg: Sort of agree, but I also know that we need to optimize to the worst case, and this is the worst case. It also makes the code simpler, and is pretty easy to change in the future.

ggggggggg:
don't merge this, the tests fail. If you merge in my other two PRs, this should become a much smaller PR, and maybe I can get it working.

ggggggggg:
So I found one problem, which is a real bug!

Line 355 of lancero_source.go in `StartRun` we're working on dropping incomplete lancero frames. We don't break out of the loop unless this condition is met. `NoHardware` never has incomplete frames, so this condition is never met. In principle this could happen with real hardware, though it seems very unlikely that reads would forever be synced  to frames without trying to do so.

```
			if err == nil && firstWord > 0 {
				bytesToRelease := 4 * firstWord
				// bytesToRelease += ((len(bytes) - 4*firstWord) / device.frameSize) * device.frameSize
				log.Printf("First frame bit at word %d, so release %d of %d bytes\n", firstWord, bytesToRelease, len(bytes))
				lan.ReleaseBytes(bytesToRelease)
				break
			}
```

Fixing this is not sufficient to get all my tests working.

Also, there is a lot of repeated code between `sampleCard` and `StartRun`, and why do we close the adapter between thee two calls?

joefowler:
I don't know whether we can afford to keep the adapter running between the two calls, if we can't guarantee that they happen quickly enough. Of course, I can't remember whether the adapter or the collector is the thing you have to stop to prevent buffer overflow.

I agree that refactoring out the shared code in `sampleCard` and `StartRun` would better fit the DRY principle. Maybe someday.

PR #27, opened 2018-06-22
File writing of LJH 2.2 can be controlled by RPC

joefowler:
Added RPC commands `WriteControl` and `WriteComment`. The latter just adds (or replaces) a `comment.txt` file to the directory where LJH files are being written. The former has requests Start, Stop, Pause, and Unpause. If the request is Start, then the additional information is used to select file type (only LJH2.2 is valid so far), number of records (value is ignored), and the base path.

File naming works like this:
* RPC command fixes the `BasePath`. I would suggest changing this only once per cool down, but user can choose other methods. Because users gonna use. Let's suppose `BasePath="/data/2018A"`
* For the first file-set written on June 22, the directory will automatically be `/data/2018A/20180622/0000`, and the file names will be `/data/2018A/20180622/0000/20180622_run0000_chan1.ljh` and similar.
* For the next file-set written that day (determined by `time.Now()`), the last directory increments to `0001` and the files are named `20180622_run0001_chan*.ljh`.
* It is an error to try to write more than 10,000 LJH file-sets in one day. You can work around this by changing `BasePath`.

Still on the to do list are:

1. Allow user to specify a total number of records to write.
1. Allow LJH3 files to be written.
1. Get ncols x nrows info for Lancero data, also perhaps "card number 1 of 3" or whatever.
1. Figure out the timestamp offset for the LJH headers. Does this header item even make sense any more now that we're devoting 64 bits to the per-pulse timestamp? It was formerly used so that we could pack _relative_ timestamps into 4 (then later, 5) bytes in the LJH 2.0 and 2.1 formats, and this was a global timestamp added to those compressed per-pulse timestamps.
1. Emit some kind of information about how much data was written. I think this should be a message on the STATUS port, either part of the new `ALIVE` heartbeat message or its own new message. It would contain a channel-by-channel `[]int` with the number of records.

I propose that we merge Galen's PR about Lancero-without-hardware (#26) before this one. But after that, I this we should merge this even with those five to-do items still outstanding. I think they represent good but not essential features. We'll see how easy they are to knock out on Monday.

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"os"
+	"strings"
 	"sync"
 	"time"
 
@@ -34,6 +36,7 @@ type DataSource interface {
 	ConfigureProjectorsBases(int, mat.Dense, mat.Dense) error
 	ChangeTriggerState(*FullTriggerState) error
 	ConfigureMixFraction(int, float64) error
+	WriteControl(*WriteControlConfig) error
 }
 
 // ConfigureMixFraction provides a default implementation for all non-lancero sources that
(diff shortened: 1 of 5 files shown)


data_source.go
```diff
@@ -103,6 +108,97 @@ func (ds *AnySource) StartRun() error {
...
+			}
+			return fmt.Sprintf("%s/%s_run%4.4d_%%s.ljh", thisDir, today, i), nil
+		}
+	}
+	return "", fmt.Errorf("out of 4-digit ID numbers for today in %s", todayDir)
+}
+
+// WriteControl changes the data writing start/stop/pause/unpause state
+func (ds *AnySource) WriteControl(config *WriteControlConfig) error {
+	if strings.HasPrefix(config.Request, "Pause") {
```
ggggggggg: should we be case insensitive here? are we elsewhere?

data_source.go
```diff
@@ -103,6 +108,97 @@ func (ds *AnySource) StartRun() error {
...
+		ds.writingState.Active = false
+		ds.writingState.Paused = false
+		ds.writingState.Filename = ""
+
+	} else if strings.HasPrefix(config.Request, "Start") {
+		if config.FileType != "LJH2.2" {
+			return fmt.Errorf("WriteControl FileType=%q, needs to be %q",
+				config.FileType, "LJH2.2")
+		}
+		filenamePattern, err := makeDirectory(config.Path)
```
ggggggggg: It looks like the client (dastard-commander) is responsible for sending the same path with each file writing request?

data_source.go
joefowler: Yeah, probably should make this case-insensitive. See next push.

data_source.go
joefowler: Yes. What else could we do?

I guess we could use the previous value whenever `config.Path` is the empty string. I'll investigate.

PR #31, opened 2018-06-26
add od -tx like output to log

ggggggggg:
pulled this out of NoHardware since I can't figure out what is going wrong there, and this is a small unrelated part that I like

### DIFF
--- a/lancero/lancero.go
+++ b/lancero/lancero.go
@@ -7,7 +7,10 @@
 package lancero
 
 import (
+	"bytes"
+	"encoding/hex"
 	"fmt"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


ggggggggg:
One of the tests pass, one has download issues.

ggggggggg:
the 1.10 test seems to have passed, the 1.9 test failed due to downloading issues

joefowler:
Those problems aren't downloading issues.  Is this another item added in Go 1.10?

```text
lancero/lancero.go:166:13: undefined: hex.NewEncoder
```

ggggggggg:
https://golang.org/doc/go1.10

Yes, `NewEncoder` is new as of 1.10.

Should we bother supporting 1.9?

joefowler:
Good question. See [Go's Release History](https://golang.org/doc/devel/release.html). It looks like they have been doing major releases nearly 2x per year since 2012. I don't know if this document sheds any light on what _we_ should do. Here's their [planned release cycle](https://github.com/golang/go/wiki/Go-Release-Cycle).

My earlier opinion was that we should support 1.9, largely because it seemed painless. But now I am becoming skeptical. Support even for security fixes in 1.9 should stop around August 2018, or whenever 1.11 comes out. Ubuntu comes with 1.10, as does Mac Ports. In fact, aside from Travis, we do not run any of Dastard on any Go 1.9 platform.

Conclusion: I vote to drop 1.9 testing and try to support only 1.10 and later. We can drop 1.10 at some point in the future...

ggggggggg:
I'm pro it. Merge this, and I can make a seperate PR that removes testing, and removes all our round workarounds.

joefowler:
Nah. I'll just push the Travis change straight to Master branch.

PR #33, opened 2018-06-26
Signed error data channels are handled properly

joefowler:
This is to properly handle the fact that some channels use `RawType` (`uint16`) for the data, but the actual contents are signed `int16` values. This must be considered in computing several things. From more to less important:

- Mixing of error into FB
- Triggers
- Decimated data streams (if you're using the averaging method)
- `AnalyzeData` analysis values like pulse average and RMS

It might be that the analysis values for error channels in the TDM system are meaningless, but the mixing is important, and triggering can be, too.

I believe this PR includes tests for all the new features, tests of mix, triggering, decimation, and analysis that would all have failed before these changes.

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -90,7 +90,8 @@ func Start(ds DataSource) error {
 type AnySource struct {
 	nchan        int      // how many channels to provide
 	chanNames    []string // one name per channel
-	sampleRate   float64  // samples per second
+	signed       []bool
+	sampleRate   float64 // samples per second
 	lastread     time.Time
 	nextFrameNum FrameIndex // frame number for the next frame we will receive
 	output       []chan DataSegment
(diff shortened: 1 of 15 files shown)


data_source.go
```diff
@@ -236,6 +238,20 @@ func (ds *AnySource) Running() bool {
 	}
 }
 
+// Signed returns a per-channel value: whether data are signed ints.
+func (ds *AnySource) Signed() []bool {
+	return make([]bool, ds.nchan)
+}
+
+// VoltsPerArb returns a per-channel value scaling raw into volts.
+func (ds *AnySource) VoltsPerArb() []float32 {
```
ggggggggg: This seems incomplete, the Abaco source isn't even going to have volts per arb, it has phase per arb.

The simulated sources also don't have volts per arb, they just have arbs.

joefowler: This is a good point, but I considered the more correct "physical-units-per-arb" to be hopelessly confusing. Let's call it Volts-per-arb for now, and we can come up with a more sensible name when the Abaco or Roach is online. Maybe `physUnitsPerArb` is totally fine, but I'll make the change when the time comes.

joefowler:
Yes, this was a big change, bigger than I expected, too. If I had decided that triggering and decimation and analysis didn't need to be correct for error channels, it could have been rather smaller. I made sure always to branch on signed/unsigned before looping over samples.

I am going to merge this soon, but I want to get microscope also working with signed data. Until I do that, I won't know that this PR is correct.

joefowler:
Oops. Don't merge yet. I see a problem....

PR #45, opened 2018-06-28
Couple triggers

### LINKED ISSUE #38: Couple FB triggers to error and vice versa
As in MATTER. This will require working through any bugs in how we handle secondary triggers.

joefowler:
Couple FB -> error and vice versa via RPC call. Fixes #38

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -422,9 +422,9 @@ func (ds *AnySource) ConfigurePulseLengths(nsamp, npre int) error {
 	return nil
 }
 
-// SetCoupling is a no-op for generic data sources
+// SetCoupling is not allowed for generic data sources
 func (ds *AnySource) SetCoupling(status CouplingStatus) error {
-	return nil
+	return fmt.Errorf("Generic data sources do not support FB/error coupling")
 }
 
 // DataSegment is a continuous, single-channel raw data buffer, plus info about (e.g.)
--- a/data_source.go
+++ b/data_source.go
@@ -40,6 +40,7 @@ type DataSource interface {
 	ChangeTriggerState(*FullTriggerState) error
 	ConfigureMixFraction(int, float64) error
 	WriteControl(*WriteControlConfig) error
+	SetCoupling(CouplingStatus) error
 }
 
 // ConfigureMixFraction provides a default implementation for all non-lancero sources that
@@ -418,6 +419,11 @@ func (ds *AnySource) ConfigurePulseLengths(nsamp, npre int) error {
 	for _, dsp := range ds.processors {
 		go dsp.ConfigurePulseLengths(nsamp, npre)
 	}
+	return nil
+}
+
+// SetCoupling is a no-op for generic data sources
+func (ds *AnySource) SetCoupling(status CouplingStatus) error {
 	return nil
 }
 
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -528,6 +528,9 @@ func (ls *LanceroSource) VoltsPerArb() []float32 {
 
 // SetCoupling set up the trigger broker to connect err->FB, FB->err, or neither
 func (ls *LanceroSource) SetCoupling(status CouplingStatus) error {
+	// Notice that status == NoCoupling will visit both else clauses in this
+	// function and therefore delete the connections from either sort of coupling.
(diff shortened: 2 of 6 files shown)


lancero_source.go
```diff
@@ -525,3 +525,27 @@ func (ls *LanceroSource) VoltsPerArb() []float32 {
 	}
 	return v
 }
+
+// SetCoupling set up the trigger broker to connect err->FB, FB->err, or neither
+func (ls *LanceroSource) SetCoupling(status CouplingStatus) error {
```
ggggggggg: I think this would be more readable with one clause for NoCoupling rather than two if/else, but it looks correct.

rpc_server.go
```diff
@@ -334,6 +334,61 @@ func (s *SourceControl) WriteComment(comment *string, reply *bool) error {
...
+
+const (
+	NoCoupling CouplingStatus = iota + 1 // FB and error aren't coupled
+	FBToErr                              // FB triggers cause secondary triggers in error channels
+	ErrToFB                              // Error triggers cause secondary triggers in FB channels
+)
+
+// CoupleErrToFB turns on or off coupling of Error -> FB
+func (s *SourceControl) CoupleErrToFB(couple *bool, reply *bool) error {
+	s.mu.Lock()
```
ggggggggg: should you check for source == nil before getting lock?

rpc_server.go
```diff
@@ -334,6 +334,61 @@ func (s *SourceControl) WriteComment(comment *string, reply *bool) error {
...
+	return err
+}
+
+// CoupleFBToErr turns on or off coupling of FB -> Error
+func (s *SourceControl) CoupleFBToErr(couple *bool, reply *bool) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.activeSource == nil {
+		return fmt.Errorf("No source is active")
+	} else if s.status.SourceName != "Lancero" {
```
ggggggggg: just make the AnySource method return an error instead of retruning nil, then you don't need to special case lancero here.

data_source.go
```diff
@@ -421,6 +422,11 @@ func (ds *AnySource) ConfigurePulseLengths(nsamp, npre int) error {
 	return nil
 }
 
+// SetCoupling is a no-op for generic data sources
+func (ds *AnySource) SetCoupling(status CouplingStatus) error {
```
ggggggggg: this should return an error like "setting coupling for AnySourc not yet implemented"

rpc_server.go
ggggggggg: Wont this have the same race condition issues as SetMix, and thus need the same solution (eg a go routine?), also the function that grabs the lock should be the same, which I think means it should be in lancero.Lancero.SetCoupling.

joefowler: At first I agreed, but no. It would be a race condition to check for `s.activeSource==nil` without the lock, because a different RPC client might request Start or Stop commands concurrently. These will change the `s.activeSource ` value. And the lock that's being held here is s's lock.

Another difference is that SetMix needs to hold the _source's_ lock to proceed, while changing group trigger needs to hold only the trigger broker's lock. The source's lock is hard to get, because `blockingRead` holds it for most of the time and then tries to grab it again as soon as it lets it go. The broker's lock, by contrast, is normally held for a small fraction of the time (only while primary triggers are being checked for secondary triggers).

I don't think LanceroSource.SetCoupling needs a lock, because it's not manipulating any `LanceroSource` variables. It's just calling a method of the trigger broker, which itself uses a lock.

Anyway, I think the issue here is confusion between the locks that lock `SourceControl` and those that lock an `AnySource`. I'm going to stand by this way of doing it.

data_source.go
joefowler: Trigger coupling is not ever going to be implemented on generic sources, because they don't have error-like "auxiliary channels" to go with each primary channel.

However, you might be right that this ought to return an error (other than not-implemented-yet). Let me think about it. Yeah, I guess that's right. Let's make this return error, thereby eliminating the special case in the RPC server.

rpc_server.go
ggggggggg: Thanks for the explanation, I buy it.

PR #47, opened 2018-06-29
File writing: fix headers for LJH files

### LINKED ISSUE #28: Finish LJH 2.2 file writer: appropriate header info
The LJH 2.2 needs to complete the header information. With PR #27 we have control over file writing, but not all of the header is complete or sensible.  Here is the header from a sim pulse source, taken last week on my Mac:

```text
#LJH Memorial File Format
Save File Format Version: 2.2.0
Software Version: Generated by DASTARD
Digitized Word Size In Bytes: 2
Presamples: 200
Total Samples: 1000
Channel: 0
Timestamp offset (s): 0.000000
Timebase: 0.000100
Number of rows: 0
Number of columns: 0
#End of Header
```

And here is the header from `20180425_B_chan33.ljh` taken on Horton.

```text
#LJH Memorial File Format
Save File Format Version: 2.2.0
Software Version: MATTER Client 1.1.3
Software Driver Version: unknown
Server Start Time: 25 Apr 2018, 20:35:04 GMT
Acquisition Mode: 0
Digitized Word Size in Bytes: 2
File First Record Time: 25 Apr 2018, 23:02:48 GMT
Channel: 33
Number of columns: 8
Number of rows: 8
Column number (from 0-7 inclusive): 2
Row number (from 0-7 inclusive): 0
Location: Blg 81
Cryostat: Horton
Bias level (% Rn): 
T set point (mK): 70
"Heater" output (%): 
Magnetic field (mg): 
Sample: 6 sided metrology
Photon source: easyXSI
Operator(s): Y. Joe, GCO
System comment: 
#End of description
Number of Digitizers: 1
Number of Active Channels: 1
Timestamp offset (s): 1524688504.806504
Digitizer: 1
Description: n/a
Master: yes
Bits: 16
Effective Bits: unknown
Anti-alias low-pass cutoff frequency (Hz): n/a
Timebase: 2.560000e-06
Number of samples per point: 1
Presamples: 1875
Total Samples: 3750
Dummy: 0
Dummy: 0
Dummy: 0
Dummy: 0
Dummy: 0
Dummy: 0
Dummy: 0
Dummy: 0
Dummy: 0
Digitizer Channel: 1.0
Dummy: 0
Range: 0.500000
Offset: 0.500000
Coupling: DC
Impedance: n/a
Inverted: No
Preamp gain: 1.0
Discrimination level: n/a
#End of Header
```

Some of these values, including the ones labeled "Dummy" are there for Igor compatibility. I am done with Igor compatibility. Mow is the time to break that--or at least we can have the Igor people fix that. But here is my to-do list for getting all sensible facts observed in the Matter-produced files also into the Dastard-produced ones:

- [x] Dastard needs to maintain a version number, and it needs to appear here.
- [x] Number of rows/columns needs to be read from Lancero, and somehow also get "Card 1 of 2" into the header.
- [x] Drop number of rows/columns from non-Lancero sources, but DO always say number of channels. Store both the internal channel numbers 0 <= chan < N, and the channel _name_.
- [x] Timestamp offset (s) needs to be filled just because this used to be such an important field. MASS might require it to exist? Might as well make it equal the "Server Start Time" as it does in this Matter example.
- [x] Add Server Start Time and File First Record Time.
- [x] Add "Number of samples per point", which is the decimation level.
- [x] Dastard needs to not set file BasePath to empty string upon starting.

joefowler:
Also uses a Makefile. That screwed up travis testing at first, but now I understand why. It is a hassle, but it's the only way I can think of to automate the `-ldflags` that bundle the git hash and build date into the built binary.

I have to say, this was way more work than I expected.

Fixes #28.


data_source.go
```diff
@@ -89,13 +87,38 @@ func Start(ds DataSource) error {
 	return nil
 }
 
+// RowColCode holds an 8-byte summary of the row-column geometry
```
joefowler: I admit that this packing of row/col number and counts is probably overkill, but it was fun!

ljh/ljh.go
```diff
@@ -52,13 +53,21 @@ type Writer struct {
 	ChanNum         int
 	Presamples      int
 	Samples         int
+	FramesPerSample int
 	Timebase        float64
-	TimestampOffset float64
```
ggggggggg: I thought we were going to drop this? Maybe just rename it something like "time writing started"

PR #51, opened 2018-06-29
handle max=min and max<min cases in TriangleSource

ggggggggg:
Before if you put max=min in the triangle source, it just sent no data, and I found it very confusing.

### DIFF
--- a/simulated_data_sources.go
+++ b/simulated_data_sources.go
@@ -38,19 +38,27 @@ func (ts *TriangleSource) Configure(config *TriangleSourceConfig) error {
 	}
 	ts.runMutex.Lock()
 	defer ts.runMutex.Unlock()
+	if config.Min > config.Max {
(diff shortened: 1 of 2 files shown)


simulated_data_sources.go
```diff
@@ -38,19 +38,27 @@ func (ts *TriangleSource) Configure(config *TriangleSourceConfig) error {
...
+	if nrise > 0 {
+		ts.cycleLen = 2 * int(nrise)
+		ts.onecycle = make([]RawType, ts.cycleLen)
+		var i RawType
+		for i = 0; i < nrise; i++ {
+			ts.onecycle[i] = config.Min + i
+			ts.onecycle[int(i)+int(nrise)] = config.Max - i
+		}
+	} else if nrise == 0 {
+		ts.cycleLen = 1
```
joefowler: Why not make this bigger, say, 10000?

ggggggggg: thats probably a good idea, i just followed the natural extension of the previous code. Maybe there should be a minimum cycle length?

PR #52, opened 2018-06-29
Lancero debug

ggggggggg:
* change `LanceroSource.Mix` from `[]Mix` to `[]*Mix` to make it modify in place and actually store `lastFb`
* fix `lancero.OdDashTX` to flip the order of every set of 4 bytes to match the output of `od -tx`
* move `acquire` to `cmd/acquire` which I think makes `go install` work
* add `cmd/oddashtx` which like acquire with no options that just prints out `od -tx` like output
* comment out some very verbose prints in `lancero` tests


lancero/cmd/acquire/.gitignore
```diff
@@ -0,0 +1 @@
+acquire
```
joefowler: Check this file. You will need to include the following lines in .gitignore:
```text
acquire
!acquire/
```

The first will ignore the binary called acquire AND all directories named acquire. Then the second says actually, don't ignore directories named acquire.

Maybe there is a more elegant way to do this, but I don't know it. (Don't build in the source directory??)

joefowler: Never mind. I misunderstood which .gitignore file this was.

PR #55, opened 2018-07-05
Dastard tracks NSAMP and weights mix fraction appropriately

joefowler:
This PR will track the Lancero card's NSAMP. For now, the user has to configure this (presumably via dastard-commander), but eventually it can be done by Cringe making an RPC call.

* accept NSAMP as part of the Lancero source configuration and report same to other clients.
* compute the voltsPerArb item for error channels as 1.0 / (NSAMP*4096) because its the sum of NSAMP values from a 12-bit ADC.
* rescale the requested mix fraction by 1/NSAMP.

Ready to pull, I believe.

### DIFF
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -130,13 +130,11 @@ func (ls *LanceroSource) Configure(config *LanceroSourceConfig) error {
 	}
 	sort.Ints(config.AvailableCards)
 
-	// Ignore nsamp if not positive
-	if config.Nsamp > 16 {
-		return fmt.Errorf("LanceroSourceConfig.Nsamp=%d but requires NSAMP<=16", config.Nsamp)
+	// Error if Nsamp not in [1,16].
+	if config.Nsamp > 16 || config.Nsamp < 1 {
+		return fmt.Errorf("LanceroSourceConfig.Nsamp=%d but requires 1<=NSAMP<=16", config.Nsamp)
 	}
-	if config.Nsamp > 0 {
-		ls.nsamp = config.Nsamp
-	}
+	ls.nsamp = config.Nsamp
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


lancero_source.go
```diff
@@ -123,6 +129,14 @@ func (ls *LanceroSource) Configure(config *LanceroSourceConfig) error {
 		config.AvailableCards = append(config.AvailableCards, k)
 	}
 	sort.Ints(config.AvailableCards)
+
+	// Ignore nsamp if not positive
```
ggggggggg: lets not do any ignore, just throw an error

lancero_source.go
```diff
@@ -167,7 +181,7 @@ func (ls *LanceroSource) ConfigureMixFraction(processorIndex int, mixFraction fl
 	go func() {
 		ls.runMutex.Lock()
 		defer ls.runMutex.Unlock()
-		ls.Mix[processorIndex].mixFraction = mixFraction
+		ls.Mix[processorIndex].mixFraction = mixFraction / float64(ls.nsamp)
```
ggggggggg: What about renaming `Mix.mixFraction` to `Mix.mixFractionOverNsamp` so we don't have the case that numbers with different value have the same name. Meaning, the user sends `MixFraction` over the RCP server, so anything called `MixFraction` should have the value they sent.

lancero_source.go
joefowler: I was thinking we want to be able to configure the Lancero without specifying NSAMP, which this accomplishes. 

However, I guess what we really want is the opposite: we eventually want Cringe to be able to set NSAMP without changing any other configurations. Of course, this can be solved by at least two other approaches:

1. Make an RPC command that sets only the NSAMP value and doesn't touch other Lancero configurations. Cringe will use that.
1. Cringe "listens" to the ZMQ status port, learns the current Lancero config, and sends back an RPC request using that config (but with the NSAMP updated, obviously).

Maybe I'll change this to an error, as you suggest.

lancero_source.go
joefowler: Hmm. I see your point. Maybe I can think of something more terse. Perhaps `scaledMixFraction` or `mixRatio` or `mixMultiplier` or better, leave "mix" out of it: `errorMultiplier`.

Okay, I'll make the proposed revisions and pull.

ggggggggg: I prefer the explicit name, it's a form of self documentation. The value is only used in a few spots, terseness won't buy you much.

PR #56, opened 2018-07-06
changing record nsamp removes projectors

### LINKED ISSUE #53: projectors should be removed when number of samples in a record changes
Otherwise dastard will panic when it tries to use projectors on a record of the wrong length.

Dastard already checks that the projectors length matches the number of samples when setting projectors.

ggggggggg:
fixes #53

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -45,8 +45,8 @@ type DataSource interface {
 
 // ConfigureMixFraction provides a default implementation for all non-lancero sources that
 // don't need the mix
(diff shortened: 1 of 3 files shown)


data_source.go
```diff
@@ -45,8 +45,8 @@ type DataSource interface {
 
 // ConfigureMixFraction provides a default implementation for all non-lancero sources that
 // don't need the mix
-func (ds *AnySource) ConfigureMixFraction(channelIndexex int, mixFraction float64) error {
```
ggggggggg: What you don't like indexex?

PR #59, opened 2018-07-06
Trigger states: restore from config file

### LINKED ISSUE #58: Trigger conditions for a newly started source are hard-coded, not restored from config
The loop in `data_source.go`, specifically in `func (ds *AnySource) PrepareRun()` sets various "sensible defaults" for the trigger conditions when a new source is started.

If the trigger settings have been saved into a viper configuration file, *these* ought to override any defaults. Also, the defaults ought to be all triggers off.

joefowler:
The trigger state was being saved to the config file but not restored from it. This is now done, substituting sensible defaults (all trigger types off) for channels not in the saved state.

Fixes #58

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -355,9 +355,16 @@ func (ds *AnySource) PrepareRun() error {
 	}
 	// Use defaultTS for any channels not in the stored state.
 	// This will be needed any time you have more channels than in the
-	// last saved configuration.
-	defaultTS := TriggerState{AutoDelay: 250 * time.Millisecond,
-		LevelLevel: 4000, EdgeLevel: 100, EdgeRising: true}
+	// last saved configuration. All trigger types are disabled.
+	defaultTS := TriggerState{
+		AutoTrigger:  false,
+		AutoDelay:    250 * time.Millisecond,
+		EdgeTrigger:  false,
+		EdgeLevel:    100,
+		EdgeRising:   true,
+		LevelTrigger: false,
+		LevelLevel:   4000,
+	}
 
 	for channelNum, dataSegmentChan := range ds.output {
 		dsp := NewDataStreamProcessor(channelNum, ds.broker)
@@ -373,10 +380,6 @@ func (ds *AnySource) PrepareRun() error {
 		}
 		dsp.TriggerState = *ts
 
-		// TODO: don't just set decimation and samples/presamples to arbitrary values
-		dsp.Decimate = false
-		dsp.DecimateLevel = 2
-		dsp.DecimateAvgMode = true
 		dsp.NPresamples = 250
 		dsp.NSamples = 1000
 
--- a/data_source.go
+++ b/data_source.go
@@ -353,7 +353,9 @@ func (ds *AnySource) PrepareRun() error {
 			}
 		}
 	}
-	// use defaultTS for any channels not in the stored state
+	// Use defaultTS for any channels not in the stored state.
+	// This will be needed any time you have more channels than in the
+	// last saved configuration.
 	defaultTS := TriggerState{AutoDelay: 250 * time.Millisecond,
 		LevelLevel: 4000, EdgeLevel: 100, EdgeRising: true}
 
--- a/data_source.go
+++ b/data_source.go
@@ -380,9 +380,6 @@ func (ds *AnySource) PrepareRun() error {
 		}
 		dsp.TriggerState = *ts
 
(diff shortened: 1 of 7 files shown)


data_source.go
```diff
@@ -337,6 +338,27 @@ func (ds *AnySource) PrepareRun() error {
...
+		for _, chnum := range ts.ChanNumbers {
+			if chnum < ds.nchan {
+				tsptrs[chnum] = &(fts[i].TriggerState)
+			}
+		}
+	}
+	// Use defaultTS for any channels not in the stored state.
+	// This will be needed any time you have more channels than in the
+	// last saved configuration.
+	defaultTS := TriggerState{AutoDelay: 250 * time.Millisecond,
```
ggggggggg: This should be all OFF! It's easy to turn them on, it's super frustrating to track down why a channel is triggering when I don't expect it to do so.

data_source.go
```diff
@@ -345,19 +367,18 @@ func (ds *AnySource) PrepareRun() error {
...
+		dsp.DecimateLevel = 2
 		dsp.DecimateAvgMode = true
-		dsp.LevelTrigger = false
-		dsp.LevelLevel = 4000
-		dsp.EdgeTrigger = true
-		dsp.EdgeLevel = 100
-		dsp.EdgeRising = true
-		dsp.NPresamples = 200
+		dsp.NPresamples = 250
 		dsp.NSamples = 1000
```
ggggggggg: Can we save NSamples and NPresamples as well? Maybe in a different PR, but a keything to minimize steps for doing the same thing every time.

data_source.go
```diff
@@ -345,19 +367,18 @@ func (ds *AnySource) PrepareRun() error {
...
+		if ts == nil {
+			ts = &defaultTS
+		}
+		dsp.TriggerState = *ts
+
+		// TODO: don't just set decimation and samples/presamples to arbitrary values
 		dsp.Decimate = false
-		dsp.DecimateLevel = 3
+		dsp.DecimateLevel = 2
 		dsp.DecimateAvgMode = true
```
ggggggggg: Why is Decimate on by default?

lancero/lancero.go
```diff
@@ -193,6 +193,7 @@ func OdDashTX(b []byte, maxLines int) string {
 	outBuffer.WriteString("L = least significant byte, M = most significant byte\n")
 	outBuffer.WriteString("framebit in fbkL\n")
 	outBuffer.WriteString("errLerrM fbkLfdbM errLerrM fbkLfdbM errLerrM fbkLfdbM errLerrM fbkLfdbM \n")
```
ggggggggg: Are these right once the bits are flipped to make words read little endian?

data_source.go
joefowler: This is OFF. But I can be more explicit, if you like.

data_source.go
joefowler: It isn't:

`dsp.Decimate = false`

data_source.go
joefowler: Actually, I think these ARE already saved. Let me figure this out...

data_source.go
ggggggggg: But then why set DecimateLevel at all? Should it just be set to the zero value, and when someone turns it on they can set the values they want.

lancero/lancero.go
joefowler: I'll check. I think they should read "fMfLeMeL"

I'll fix.

data_source.go
```diff
@@ -355,9 +355,16 @@ func (ds *AnySource) PrepareRun() error {
...
+	// last saved configuration. All trigger types are disabled.
+	defaultTS := TriggerState{
+		AutoTrigger:  false,
+		AutoDelay:    250 * time.Millisecond,
+		EdgeTrigger:  false,
+		EdgeLevel:    100,
+		EdgeRising:   true,
+		LevelTrigger: false,
+		LevelLevel:   4000,
+	}
```
ggggggggg: I think this line should read `defaultTS := TriggetState{}`, I don't see why we should have any default nonzero values. What effect does this have? I guess it effectively sets the text in the dastard commander trigger ui?

It's good enough that I'll stop complaining about it and let you decide.

joefowler: Right: These other values set the text that you'll see in the GUI when you start to use it. They ensure that no nonsensical values appear there.

PR #62, opened 2018-07-09
publish number of records written

ggggggggg:
Uses a single counter for all types of records, resets the count any time any file writing is turned on or off.

This adds a 2nd synchronization point to dastard after record publication. The other is during group triggering.


ggggggggg:
thanks

PR #64, opened 2018-07-11
Fix lancero timing

### LINKED ISSUE #39: Timestamps might be more irregular than they need to be
Auto triggering at 250 ms suggests that either the auto-trigger algorithm is not regular, or (more likely) that the time stamping algorithm is wrong. Is there a long latency, or is the wrong time being assigned? Or is there an error in the computation of the first sample's timestamp from the last sample's timestamp?

joefowler:
Commit e02358f fixes #39 this in grand fashion for the Lancero source. In my test case, there are 1375 records taken with 250 ms pulse auto-triggers, so over 5 minutes of data.  Every record is 34722 frames apart, as expected (7.2 µs timebase, so 34722 frames = 249.9984 ms).

But the cool part is that the mean time between records' timestamps is 250.003 ms with a standard deviation of 3.77 µs. Hooray!

While the standard deviation for simulated pulse sources is more like 50 µs, we also don't really care.

One problem with this PR is that it doesn't have a test to exercise the bug. Unfortunately, the bug lived in `func (ls *LanceroSource) distributeData()`, and I don't know how to test that without actual hardware. Could something go in `TestNoHardwareSource` ?

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -114,14 +114,15 @@ func rcCode(row, col, rows, cols int) RowColCode {
 // AnySource implements features common to any object that implements
 // DataSource, including the output channels and the abort channel.
 type AnySource struct {
-	nchan        int          // how many channels to provide
-	name         string       // what kind of source is this?
-	chanNames    []string     // one name per channel
-	chanNumbers  []int        // names have format "prefixNumber", this is the number
-	rowColCodes  []RowColCode // one RowColCode per channel
-	signed       []bool       // is the raw data signed, one per channel
-	voltsPerArb  []float32    // the physical units per arb, one per channel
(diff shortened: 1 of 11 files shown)


joefowler:
No, the problem was with the fact that I was creating DataSegment objects with the default null value for their `framePeriod` field. That, in turn, was causing the time arithmetic to be incorrect by an amount of time equal to however many frames were in the DataStream to which the new DataSegment was being appended. Since that was of order 10 ms of data, we were getting that scale or errors.

The extra output of the Wait function was not actually enough to fix the problem. Still, I left it in because it is...

<this is a comment I started to write 2 days ago and never finished. Here it is for the record.>

PR #65, opened 2018-07-11
Off file writing

ggggggggg:
Implements OFF file writing, and changes the WriteControl RPC call API to support it.

My biggest remaining concern is how to deal with the case of someone asking to `WriteOFF` but they haven't loaded projectors. See comment in `data_source.go`. What will happen if there is an error during file writing, will dastard commander know about it?

Replaces #9


ljh/ljh.go
```diff
@@ -309,12 +309,14 @@ func (w *Writer3) WriteRecord(firstRisingSample int32, rowcount int64, timestamp
 
 // Flush flushes buffered data to disk
 func (w Writer3) Flush() {
-	w.writer.Flush()
+	if w.writer != nil {
+		w.writer.Flush()
+	}
```
joefowler: Good. This was a problem with the LJH 2 writer, too, right?

off/off.go
```diff
@@ -0,0 +1,202 @@
+// Package off provides classes that write OFF files
+// OFF files store TES pulses projected into a linear basis
+// OFF files have a JSON header followed by a single newline
+// after the header records are written sequentially in little endian format
+// bytes		type			meaning
+// 0-3      int32     recordSamples (could be calculated from nearest neighbor pulses in princple)
+// 4-7      int32     recordPreSamples (could be calculated from nearest neighbor pulses in princple)
```
joefowler: I don't understand this or the previous misspelled parenthetical. How could the number of samples or presamples be computed in principle? I mean yes it can, but that presumes a certain set of rules for how the data are split into records. I think those rules should be treated as a plan that can be changed in Dastard from time to time, maybe even with user-settable parameters. In that event, we'll want the OFF file to record these values. (Yes, you could put into the header enough information to work out the rules, but ugh!)

off/off.go
```diff
@@ -0,0 +1,202 @@
...
+// 8-15     int64     framecount
+// 16-23    int64     timestamp from time.Time.UnixNano()
+// 24-27    float32   pretriggerMean (from raw data, not from modeled pulse, really shouldn't be neccesary, just in case for now!)
+// 28-31    float32   residualStdDev (in raw data space, not Mahalanobis distance)
```
joefowler: This residual is simpler, but I dream someday of computing Mahalanobis distance. Will the header contain a version number to permit changes in this particular OFF format? And anyway, aren't you going to wish we had used [gogetdata](https://github.com/joefowler/gogetdata) pretty soon?

off/off.go
```diff
@@ -0,0 +1,202 @@
...
+	return nil
+}
+
+// WriteRecord writes a record to the file
+func (w *Writer) WriteRecord(recordSamples int32, recordPreSamples int32, framecount int64,
+	timestamp int64, pretriggerMean float32, residualStdDev float32, data []float32) error {
+	if len(data) != w.NumberOfBases {
+		return fmt.Errorf("wrong number of bases, have %v, want %v", len(data), w.NumberOfBases)
+	}
+	if _, err := w.writer.Write(getbytes.FromInt32(int32(recordSamples))); err != nil {
```
joefowler: I'm not saying I know how to improve on this myself for sure, but this seems to have too many `Write` calls and error checks. Yes, yes, premature optimization. But can't we string together the various `[]byte` slices returned by `get bytes.FromXYZ` with `append` and make one `Write` and one error check?

ljh/ljh.go
ggggggggg: Yes, I'm surprised it doesn't seem to be a problem with `w.file` as well.

off/off.go
ggggggggg: I'm assuming that you're using edgeMulti triggering with variable length records, and you could calculate the frames to previous pulse and frames to next pulse. But year, for now we'll include it. Really I should probably just put this text in the header.

off/off.go
ggggggggg: I'm using a buffered writer, I think it's equivalent to doing all that myself, except it works across calls to `WriteRecord`

ggggggggg:
can you look at https://github.com/usnistgov/dastard/pull/65/commits/75988d97341000f4eb25bd057c1397a3068a6463 ? I think it avoids a race condition, thought it doesn't acheive my goal of making every channel have the exact same number of records written when auto triggering.

joefowler:
Yeah, I cannot review that commit without eating lunch first. Sorry.

PR #68, opened 2018-07-11
Change how we lock the AnySource.WriteControl method

### LINKED ISSUE #67: RPC WriteControl does not finish!
See 08f57b4

joefowler:
This fixes some problems I was having with start/comment/stop writing files. I think that file writing was broken in the previous PR? 

I have changed how you handle the WriteConfig.

This work looks small but it took 100% of my time all afternoon, for reasons I don't really understand.

Fixes #67

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -40,7 +40,7 @@ type DataSource interface {
 	ConfigureProjectorsBases(int, mat.Dense, mat.Dense) error
 	ChangeTriggerState(*FullTriggerState) error
 	ConfigureMixFraction(int, float64) error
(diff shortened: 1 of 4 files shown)


ggggggggg:
Thanks. It worked in my tests, not sure why. While this race condition stuff is hard in go, its still way easier than in C++ so we've got that going for us.

ggggggggg:
I think this causes the potential for a hang when starting file writing, not sure why yet. But my version seems to work for me, and this one doesn't.  I see the hang with a SimPulse source with 50 channels and 100k samples/second but not with 4 or 10 channels at the same rate.

PR #70, opened 2018-07-12
name off files *.off instead of *.ljh, same for ljh3

ggggggggg:


ggggggggg:
doesn't seem to work yet, wait for a fix

PR #72, opened 2018-07-12
Synchronous source control

ggggggggg:
change the rpc server to call `ServeRequest` in a loop, so that we handle each request in turn. Remove the lock from `SourceControl`

does some minor cleanup, including removing the `Outputs` function that was only used for testing (see commit comment)

### DIFF
--- a/cmd/dastard/dastard.go
+++ b/cmd/dastard/dastard.go
@@ -93,5 +93,6 @@ func main() {
 	}
 
 	go dastard.RunClientUpdater(dastard.Ports.Status)
-	dastard.RunRPCServer(dastard.Ports.RPC)
+	dastard.RunRPCServer(dastard.Ports.RPC, true)
+
 }
--- a/cmd/dastard/dastard.go
+++ b/cmd/dastard/dastard.go
@@ -88,7 +88,7 @@ func main() {
 
 	// Find config file, creating it if needed, and read it.
 	if err := setupViper(); err != nil {
-		panic(err)
+		log.Fatal(err)
 	}
 
 	go dastard.RunClientUpdater(dastard.Ports.Status)
--- a/cmd/dastard/dastard.go
+++ b/cmd/dastard/dastard.go
@@ -3,6 +3,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"os/user"
 	"strings"
--- a/data_source.go
+++ b/data_source.go
@@ -28,7 +28,6 @@ type DataSource interface {
 	Stop() error
 	Running() bool
 	blockingRead() error
-	//Outputs() []chan DataSegment
 	CloseOutputs()
 	Nchan() int
 	Signed() []bool
@@ -472,14 +471,6 @@ func (ds *AnySource) Stop() error {
 	ds.publishSync.Stop()
 	return nil
 }
-
-// // Outputs returns the slice of channels that carry buffers of data for downstream processing.
-// func (ds *AnySource) Outputs() []chan DataSegment {
-// 	// Don't run this if PrepareRun or other sensitive sections are running
-// 	ds.runMutex.Lock()
-// 	defer ds.runMutex.Unlock()
-// 	return ds.output
-// }
 
 // CloseOutputs closes all channels that carry buffers of data for downstream processing.
 func (ds *AnySource) CloseOutputs() {
--- a/data_source.go
+++ b/data_source.go
@@ -28,7 +28,7 @@ type DataSource interface {
 	Stop() error
 	Running() bool
 	blockingRead() error
-	Outputs() []chan DataSegment
+	//Outputs() []chan DataSegment
 	CloseOutputs()
 	Nchan() int
 	Signed() []bool
@@ -473,13 +473,13 @@ func (ds *AnySource) Stop() error {
 	return nil
 }
 
-// Outputs returns the slice of channels that carry buffers of data for downstream processing.
-func (ds *AnySource) Outputs() []chan DataSegment {
-	// Don't run this if PrepareRun or other sensitive sections are running
-	ds.runMutex.Lock()
-	defer ds.runMutex.Unlock()
-	return ds.output
-}
+// // Outputs returns the slice of channels that carry buffers of data for downstream processing.
+// func (ds *AnySource) Outputs() []chan DataSegment {
+// 	// Don't run this if PrepareRun or other sensitive sections are running
+// 	ds.runMutex.Lock()
+// 	defer ds.runMutex.Unlock()
+// 	return ds.output
+// }
 
 // CloseOutputs closes all channels that carry buffers of data for downstream processing.
 func (ds *AnySource) CloseOutputs() {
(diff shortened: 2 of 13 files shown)


ggggggggg:
don't merge yet, there are still race conditions caused by this change according to `go test -race`

joefowler:
Wait, if we have a nearly infinite number of messages on `go test -race`, then how do you know which ones were caused by this branch?

ggggggggg:
I commented out the write control stuff and tested

ggggggggg:
this actually seems to work. It passed `go test -race` but has some dangling PUB sockets, which I think we decided to not worry about for now.

ggggggggg:
Note, it only passed `go test -race` because all the `WriteControl` tests are commented out.

joefowler:
Yeah, we definitely decided not to worry about dangling PUB sockets. My only concern is that they could be a problem in event of a panic or a `log.Fatal`, but let's assume they aren't until proven otherwise.

ggggggggg:
I think the future goal would have the core data loop look something like:

```go
	go func() {
		for {
			if err := ds.blockingRead(); err == io.EOF {
				break
			} else if err != nil {
				log.Fatal("blockingRead returns Error: %s\n", err.Error())
			}
			ds.blockUntilDataProcessed() // done once all data from previous read is processed, maybe using PublishSync here?
			select {
				// handle a single configuration change per blocking read
				// or maybe loop with a very short timeout to drain all channels?
				case triggerConfig := <-ds.triggerConfigChan:
					ds.ConfigureTriggers(triggerConfig)
				case mixConfig := <-ds.mixConfigChan
					ds.ConfigureMix(mixConfig)
				case writeConfig := <- ds.writeConfigChan
					ds.ConfigureWrite(writeConfig)
				default:
			}
		}
		ds.CloseOutputs()
	}()
```

joefowler:
Incidentally, on the 687raven (linux) computer, this branch passes `go test -race` 3 times out of 5. A representative race found on the 2 failures is:

```text
==================
WARNING: DATA RACE
Read at 0x00c4202300d8 by goroutine 115:
  github.com/usnistgov/dastard.(*SourceControl).broadcastHeartbeat()
      /home/pcuser/go/src/github.com/usnistgov/dastard/rpc_server.go:365 +0x55

Previous write at 0x00c4202300d8 by goroutine 95:
  ??()
      -:0 +0xffffffffffffffff
  encoding/json.(*decodeState).scanWhile()
      /usr/lib/go-1.10/src/encoding/json/decode.go:358 +0x1b7
  encoding/json.(*decodeState).literal()
      /usr/lib/go-1.10/src/encoding/json/decode.go:814 +0x65
  encoding/json.(*decodeState).value()
      /usr/lib/go-1.10/src/encoding/json/decode.go:411 +0x186
  encoding/json.(*decodeState).object()
      /usr/lib/go-1.10/src/encoding/json/decode.go:754 +0x173a
  encoding/json.(*decodeState).value()
      /usr/lib/go-1.10/src/encoding/json/decode.go:408 +0x149
  encoding/json.(*decodeState).unmarshal()
      /usr/lib/go-1.10/src/encoding/json/decode.go:189 +0x2b7
  encoding/json.(*Decoder).Decode()
      /usr/lib/go-1.10/src/encoding/json/stream.go:73 +0x36c
  net/rpc/jsonrpc.(*serverCodec).ReadRequestHeader()
      /usr/lib/go-1.10/src/net/rpc/jsonrpc/server.go:66 +0x11f
  net/rpc.(*Server).readRequestHeader()
      /usr/lib/go-1.10/src/net/rpc/server.go:589 +0x77
  net/rpc.(*Server).readRequest()
      /usr/lib/go-1.10/src/net/rpc/server.go:549 +0x6a
  net/rpc.(*Server).ServeRequest()
      /usr/lib/go-1.10/src/net/rpc/server.go:492 +0x84
  github.com/usnistgov/dastard.RunRPCServer.func2.1()
      /home/pcuser/go/src/github.com/usnistgov/dastard/rpc_server.go:482 +0x3af

Goroutine 115 (running) created at:
  github.com/usnistgov/dastard.RunRPCServer.func1()
      /home/pcuser/go/src/github.com/usnistgov/dastard/rpc_server.go:452 +0x20f

Goroutine 95 (finished) created at:
  github.com/usnistgov/dastard.RunRPCServer.func2()
      /home/pcuser/go/src/github.com/usnistgov/dastard/rpc_server.go:478 +0x4be
==================
Found 1 data race(s)
E: 18-07-12 16:38:02 [28675]dangling 'PUB' socket created at �j:174
E: 18-07-12 16:38:02 [28675]dangling 'PUB' socket created at �:174
E: 18-07-12 16:38:02 [28675]dangling 'PUB' socket created at tcp://0.0.0.0:33304:174
exit status 66
FAIL	github.com/usnistgov/dastard	2.663s
```

ggggggggg:
I think I found the problem, a remnant `go s.broadcastHeartbeat`.

ggggggggg:
I will be out totally Friday and Monday, FYI. If you do merge this, remember that the `WriteControl` tests are commented out in case you do anything where they matter.

ggggggggg:
An alternate pattern that still uses a lock, but I think would be better at working with `WriteControl` than `runMutex` is:

```go
	go func() {
		for {
			ds.configMutex.Lock()
			if err := ds.blockingRead(); err == io.EOF {
				break
			} else if err != nil {
				log.Fatal("blockingRead returns Error: %s\n", err.Error())
			}
			ds.blockUntilDataProcessed() // done once all data from previous read is processed, maybe using PublishSync here?
			ds.configMutex.Unlock()
			}
			ds.blockUntilGoodTimeToRead()
		ds.CloseOutputs()
	}()
```

Also, if I have time (I don't, but I might act like I do), I want to review each use of `runMutex` to make sure it is needed.

joefowler:
Here's a simple variation that would process all available commands before returning to `blockingRead`

```go
go func() {
	for {
		if err := ds.blockingRead(); err == io.EOF {
			break
		} else if err != nil {
			log.Fatal("blockingRead returns Error: %s\n", err.Error())
		}
		ds.blockUntilDataProcessed() 
        // done once all data from previous read is processed;
        // maybe PublishSync can send a sync message to a channel here?

		// A simple way to drain the channels that contain change-config commands
		// before proceeding to the top of the main loop and another blockingRead
		for {
			select {
				// handle a single configuration change per blocking read
				// or maybe loop with a very short timeout to drain all channels?
				case triggerConfig := <-ds.triggerConfigChan:
					ds.ConfigureTriggers(triggerConfig)
				case mixConfig := <-ds.mixConfigChan
					ds.ConfigureMix(mixConfig)
				case writeConfig := <- ds.writeConfigChan
					ds.ConfigureWrite(writeConfig)
				default:
	                break
			}
		}
	}
	ds.CloseOutputs()
}()
```

But there's a problem here: each channel will block-on-write when the RPC server tries to cause a state change until the select above reads from the channel, or the channels will need some non-zero channel capacity. How do we set those capacities? How much is enough? By picking a value, we are effectively telling the RPC server that only so many requests of each type can be queued at once.

Anyway, good work. I'll merge.

Issue #74, opened 2018-07-17
Dastard sends an empty message with topic CURRENTTIME


ggggggggg:
Is this intended?

joefowler:
I don't think so. I'll look into it. I think that was supposed to go in the settings _file_ only but not to be broadcast to clients.

ggggggggg:
I don't see it anymore in #82, though I'm not sure where it was fixed.

joefowler:
I kind of remember thinking that I fixed it, so I'm not surprised. I just couldn't find where, either.

PR #82, opened 2018-07-25
Improve projectors load

ggggggggg:


client_updater.go
```diff
@@ -113,7 +113,6 @@ var nosaveMessages = map[string]struct{}{
 // saveState stores server configuration to the standard config file.
 func saveState(lastMessages map[string]interface{}) {
 
-	lastMessages["CURRENTTIME"] = time.Now().Format(time.UnixDate)
```
joefowler: Is this what fixed the CURRENTTIME problem in #74? I liked having current time in the config.yaml file. I'd rather you didn't take it out.

Can you add it back in and see whether #74 recurs?

rpc_server.go
```diff
@@ -372,6 +374,9 @@ func (s *SourceControl) broadcastHeartbeat() {
 }
 
 func (s *SourceControl) broadcastStatus() {
+	if s.activeSource != nil {
```
joefowler: I'm not excited to have this in the STATUS object. It's fine to have it in the message to clients, but when stored in `config.yaml`, you have a problem: it's not a state that can become true just by loading the configuration file. How is that being handled?

ggggggggg: I agree with the principle that the yaml file ideally wouldn't contain things that can't be made True, but I don't think I should fix it here. This works fine, the line immediatley after loading the status from the yaml file is to `broadcastStatus` which overwrites `ChannelsWithProjectors`. There are other things that end up in the YAML file that fall into this catogory like `numberwritten`, `lancero.availablecards`, `countsseen`, `writing` (it could be made true, but it shouldn't be).

client_updater.go
ggggggggg: I added it back in, it doesn't send recreate the problem. I'll leave it in.

ggggggggg:
The failures are unrelated, I'm tempted to merge anyway since tests pass locally. Do you want to run tests on your machine first?

joefowler:
Yes, the tests pass here on my computer, too. Feel free to merge in spite of Travis.

Issue #77, opened 2018-07-20
log.Fatal doesn't print out errors on test failures


ggggggggg:
This makes debugging test failures super annoying, should we go back to `panic`?

joefowler:
What does panic do? Print a full stack track of all running goroutines? If so, I guess that's fine. I have never much cared about the `log` package in the first place.

ggggggggg:
fixed by #86

PR #86, opened 2018-07-26
Notice lancero problem

ggggggggg:
This pull request allows a source to stop due to an internal error, and adds checking for row number, column number and lsync to the lancero source. It may be too strict, but will be easy to modify it that i the case.

It also adds `ErroringSource` to test this behavior.

In the process I found that our tests could fail due to a `PublishSync` failure, and fixed those cases with `defer source.Stop()` in a few places.

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -76,12 +76,16 @@ func Start(ds DataSource) error {
 	go func() {
 		for {
 			if err := ds.blockingRead(); err == io.EOF {
-				break
+				// EOF error should occur after abortSelf has been closed
+				// ds.CloseOutputs() // why is this here, ds.Stop also calls CloseOutputs
+				return
 			} else if err != nil {
-				log.Fatal(fmt.Sprintf("blockingRead returns Error: %s\n", err.Error()))
+				// other errors indicate a problem with source, need to close down
+				fmt.Printf("blockingRead returns Error, stopping source: %s\n", err.Error())
+				ds.Stop() // will cause next call to ds.blockingRead to return io.EOF
 			}
 		}
-		ds.CloseOutputs()
+
 	}()
 	return nil
 }
@@ -464,12 +468,16 @@ func (ds *AnySource) PrepareRun() error {
 
 // Stop ends the data supply.
 func (ds *AnySource) Stop() error {
-	if ds.Running() {
-		close(ds.abortSelf)
+	if ds.abortSelf != nil {
+		select {
+		case <-ds.abortSelf:
+		default:
+			close(ds.abortSelf)
+		}
 	}
-	ds.runDone.Wait()
 	ds.broker.Stop()
 	ds.publishSync.Stop()
+	ds.CloseOutputs()
 	return nil
 }
 
@@ -477,7 +485,6 @@ func (ds *AnySource) Stop() error {
 func (ds *AnySource) CloseOutputs() {
 	ds.runMutex.Lock()
 	defer ds.runMutex.Unlock()
-
 	for _, ch := range ds.output {
 		close(ch)
 	}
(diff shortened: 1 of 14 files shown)


data_source.go
```diff
@@ -477,20 +490,20 @@ func (ds *AnySource) PrepareRun() error {
 
 // Stop ends the data supply.
 func (ds *AnySource) Stop() error {
-	if ds.Running() {
-		close(ds.abortSelf)
+	if !ds.Running() {
+		return fmt.Errorf("Anysource not running, cannot stop")
 	}
-	ds.runDone.Wait()
```
joefowler: Why isn't the new code calling `ds.Wait()` ?

lancero_source.go
```diff
@@ -468,13 +475,29 @@ func (ls *LanceroSource) distributeData() {
...
+		if ncols != dev.ncols || nrows != dev.nrows || lsync != dev.lsync || framesUsed <= 0 {
+			fmt.Printf("have ncols %v, nrows %v, lsync %v, framesUsed %v. had ncols %v, nrows %v, lsync %v, blockingReadCount %v\n",
+				ncols, nrows, lsync, framesUsed, dev.ncols, dev.nrows, dev.lsync, ls.blockingReadCount)
+			if ls.blockingReadCount > 1 {
+				// lsync calculation fails on first few reads so don't error
+				return fmt.Errorf("have ncols %v, nrows %v, lsync %v, framesUsed %v. had ncols %v, nrows %v, lsync %v",
+					ncols, nrows, lsync, framesUsed, dev.ncols, dev.nrows, dev.lsync)
+			}
+		}
+	}
```
joefowler: This block looks expensive. Is it? Did you profile? Also, it looks like it is repeated from somewhere.

joefowler: Also, is it really a good idea, given that LSYNC is computed from timing between reads, which is susceptible to errors when latency is large?

rpc_server.go
```diff
@@ -261,14 +267,28 @@ func (s *SourceControl) Stop(dummy *string, reply *bool) error {
...
-	s.status.Running = false
-	s.activeSource = nil
+	s.observeSourceStop()
 	*reply = true
 	s.broadcastStatus()
 	*reply = true
 	return nil
 }
 
+// observeSourceStop modifies s to make future messag
```
joefowler: Is this comment complete? Well, obviously it isn't.

joefowler:
Note also that this fixes #77 (or it seems that way to me).

data_source.go
ggggggggg: I'm not sure what `ds.Wait()` is, I couldn't find the definition, but the tests panic when I use it.

lancero_source.go
ggggggggg: I have run a lancero source with this block in place and it worked at 200 MB/s. I haven't profiled it, but I don't think it is expensive unless it is going to fail anyway. `FindFrameBits` only scans through two frames if it is succesful. I am a bit worried we will occsionally get lsync wrong due to random timing variation, and in that case I plan to add a `succesiveErrorsCount` or similar.

data_source.go
joefowler: It's a `sync.WaitGroup`. It lets something wait on multiple events to happen, potentially in separate goroutines. Not sure whether we need it here.

ggggggggg: `runeDone` is the `sync.WaitGroup` and is what I'm waiting on here. I think `ds.Wait()` is actually the function I'm defining here, and I was causing a stack overflow to get those panics.

rpc_server.go
ggggggggg: fixed

PR #87, opened 2018-07-26
Experiment state labels

ggggggggg:
Depends on #86, aka will get less messy once #86 is merged.

Adds two APIs for telling dastard an experiement state label
1. WriteControl with PAUSE and UNPAUSE LABEL (for higher latency applications, forces pairings of label start and label stop)
2. SetExperimentStateLabel (for low latency)

We can revisit the topic of forced start-stop pairings in the future, but this should hopefully work for raven.

When labels A,B and C are sent writes a files that looks like
```
#unixnano, label
324234242 A
342342424 B
929349239 C
```

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -280,7 +280,7 @@ func (ds *AnySource) WriteControl(config *WriteControlConfig) error {
 		}
 		ds.writingState.Active = false
 		ds.writingState.Paused = false
-		ds.writingState.filenamePattern = ""
+		ds.writingState.FilenamePattern = ""
 		if ds.writingState.experimentStateFile != nil {
 			ds.writingState.experimentStateFile.Close()
 		}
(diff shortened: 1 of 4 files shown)


joefowler:
I don't see why we want two APIs for the same basic activity. Is it just that you changed your mind halfway through and decided that each has its own use case? (I mean, you didn't change your mind so much as BAE dropped the bomb that the the labels would only last 1 second apiece, so the uncontrolled latency of your original plan was too much.)

Could you better explain how the fast usage is supposed to work for RAVEN? Does the RPC client call `SetExperimentStateLabel ` with argument `{Label:"12345"}` to start serial number 12345, and then again with `{Label:""}` to end it? Will Label="" be the special value to signify end-of-serial 12345?

ggggggggg:
I like the idea of the PAUSE,UNPAUSE API for something like SSRL 10-1 where I know it takes a while to change their monochromator settings. But mostly it was the 1 second surprise half way through.
 
This API doesn't restrict the answer to your question. A couple of options:
1. Start: `{Label: "1"}`, Stop: `{Label: "moving"}`, Start Next: `{Label: "2"}`
2: Start: `{Label: "1"}`, Stop: `{Label: "stop 1"}`, Start Next: `{Label: "2"}`
3: Start: `{Label: "1"}`, Stop: `{Label: ""}`, Start Next: `{Label: "2"}`

Ok, so really 1 and 3 are basically identical. I was thinking of 2, but I actually think I like it the least of these.

Issue #97, opened 2018-08-13
dastard gets lsync wrong a lot


ggggggggg:
We should probalby use more buffers for finding lsync, maybe append them? Also we could check to see that we get the same answer for 3 buffers in a row or something.

PR #99, opened 2018-08-14
Get lsync right attempt

ggggggggg:
- refactor how lsync is calculated. 
- Uses the timeFix from `AvailableBuffers` instead of waittime from `Wait`
- collects 200 ms worth of info for calculating lsync
- stops using more memory after ncol/nrow is found
- tested on 687raven i got lsync right every time until the lancero card stopped working an I needed to reboot (10 times approx), seems like an improvement to me

- adds some progress towards an AutoRestart button for lancero source

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -48,6 +48,7 @@ type DataSource interface {
 	ProcessSegments() error
 	RunDoneAdd()
 	RunDoneDone()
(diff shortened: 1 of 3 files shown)


joefowler:
Fixes #97

Issue #94, opened 2018-08-02
dastard writes framecount to the rowcount field in ljh22


ggggggggg:
rowcount only makes sense for a lancero source, `rowcount = framecount*nrow+row`

joefowler:
Did your comment above mean _rowcount_ only makes sense for a Lancero source?

Would you like to switch LJH2.2 to storing frame counts? Well, no switch is required, as that's what it already does. But I mean we'd change the text in the header AND update MASS to deal with it.

Or do you want to store row count, as with LJH 2.1?

ggggggggg:
Yes, I mean rowcount only makes sense for a lancero source. edited.

I can't decide easily if it should be framecount or rowcount in LJH2.1. I think I prefer rowcount because then less total code needs to change, and it will match the numbers for an external trigger if we ever get to that. But it does seem odd to write something that only makes sense for some sources. At least rowcount falls back to equal framecount when nrow = 1.

ggggggggg:
fixed by #102

PR #102, opened 2018-08-15
set pulse lengths after starting source, fixes #78

### LINKED ISSUE #78: Record length not restored correctly
I think we still have the record length of a `DataStreamProcessor` hardcoded somewhere, so that when I set the record length short (say 400 samples, 66 presamples), then quit and restart dastard, the record length is actually 256 samples. Though dastard commander says the right value, so clearly we had access to it.

ggggggggg:


ggggggggg:
The 2nd commit fixes #94

ljh/ljh.go
```diff
@@ -208,10 +208,11 @@ func (w Writer) Close() {
 // rowcount is framecount*number_of_rows+row_number for TDM or TDM-like (eg CDM) data,
 // rowcount=framecount for uMux data
 // return error if data is wrong length (w.Samples is correct length)
-func (w *Writer) WriteRecord(rowcount int64, timestamp int64, data []uint16) error {
+func (w *Writer) WriteRecord(framecount int64, timestamp int64, data []uint16) error {
 	if len(data) != w.Samples {
 		return fmt.Errorf("ljh incorrect number of samples, have %v, want %v", len(data), w.Samples)
 	}
+	rowcount := framecount*int64(w.NumberOfRows) + int64(w.RowNum)
```
joefowler: Does this mean that LJH 2.1 will contain row-count and 2.2 or 3.0 will contain frame-count?

I think I can live with that.

PR #113, opened 2018-10-08
Mix requests can handle multiple channels and generate a message to clients

### LINKED ISSUE #83: MixFraction should be reported by server
In light of the comment about not wanting to save ChannelsWithProjectors, I thought of other things that the server should expose but we shouldn't save. MixFraction is one. We could add a new message with a name like "DETAILS" that isn't saved by viper. Then we could put stuff like this in it.

joefowler:
Replaces one-RPC-per-channel Mix changes with a request that has an array of channels and a matching array of mix fractions. Dastard-commander has to be updated, too. Fixes #83

### DIFF
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -46,7 +46,7 @@ type LanceroSource struct {
 	active            []*LanceroDevice
 	chan2readoutOrder []int
 	Mix               []*Mix
-	blockingReadCount int
+	dataBlockCount    int
 	buffersChan       chan BuffersChanType
 	readPeriod        time.Duration
 	mixRequests       chan *MixFractionObject
@@ -60,11 +60,9 @@ func NewLanceroSource() (*LanceroSource, error) {
 	source.name = "Lancero"
 	source.nsamp = 1
 	source.devices = make(map[int]*LanceroDevice)
-	source.buffersChan = make(chan BuffersChanType, 100)
-	source.readPeriod = 50 * time.Millisecond
 
 	devnums, err := lancero.EnumerateLanceroDevices()
-	if len(devnums) == 0 || err != nil {
+	if err != nil {
 		return source, err
 	}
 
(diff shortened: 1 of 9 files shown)


lancero_source.go
```diff
@@ -569,7 +577,18 @@ func (ls *LanceroSource) getNextBlock() chan *dataBlock {
 				panic(fmt.Sprintf("timeout, no data from lancero after %v / %v", panicTime, ls.readPeriod))
 
 			case mfo := <-ls.mixRequests:
-				ls.Mix[mfo.ChannelIndex].errorScale = mfo.MixFraction / float64(ls.nsamp)
+				for i, index := range mfo.ChannelIndices {
+					if index < 0 || index >= len(ls.Mix) {
+						continue
+					}
```
ggggggggg: Don't silently ignore incorrect input, throw an error, or panic. Also it was checked in `ConfigureMixFraction`. I would just removed this check and let it panic on out of bounds.

rpc_server.go
```diff
@@ -156,8 +156,11 @@ type MixFractionObject struct {
 // But changes to the mix settings need to be kept separate from LanceroSource.distrubuteData,
 // which is part of the data-*production* step, not the data-processing step.
 func (s *SourceControl) ConfigureMixFraction(mfo *MixFractionObject, reply *bool) error {
-	err := s.ActiveSource.ConfigureMixFraction(mfo)
+	currentMix, err := s.ActiveSource.ConfigureMixFraction(mfo)
 	*reply = (err == nil)
+	if err == nil {
```
ggggggggg: make broadcast unconditional?

joefowler: Sure. I'll take care of that and then finish this PR. Thanks for reviewing.

lancero_source.go
```diff
@@ -569,7 +577,15 @@ func (ls *LanceroSource) getNextBlock() chan *dataBlock {
 				panic(fmt.Sprintf("timeout, no data from lancero after %v / %v", panicTime, ls.readPeriod))
 
 			case mfo := <-ls.mixRequests:
-				ls.Mix[mfo.ChannelIndex].errorScale = mfo.MixFraction / float64(ls.nsamp)
+				for i, index := range mfo.ChannelIndices {
+					fraction := mfo.MixFractions[i]
```
ggggggggg: It deleted my other comment because I just changed this in the code. But I removed the bounds check with silent ignore here. We should just panic on out of bounds and fix the source. Also the bounds are pre-checked in ConfigureMixFaction.

joefowler: I think it's fine to skip checks here, because--as you say--they are already checked in `ConfigureMixFraction`.

PR #138, opened 2019-04-16
Read lsync, nrows, nsamp, clockMHz from cringeGlobals.json

ggggggggg:
This changes `LanceroSource.Configure` to read from `~/.cringe/cringeGlobals.json` to learn `lsync`, `nrows`, `Nsamp`, and `clockMHz`. In `sampleCard` where we calculate `nrows` and `lsync` there is now an error if `nrows` doesn't match, and a warning if `lsync` doesn't match. 

Tests that used `LanceroSource.Configure` to set these values now have to reach inside of `LanceroSource` to set these values directly, so they are a bit uglier. Also for testing, it reads from `lancero/test_data/cringeGlobals.json` since `~/.cringe/cringeGlobals.json` won't exist on travis or my laptop.

I think this will work without any modifications to dastard commander, since Dastard will just ignore extraneous values in the json that gets sent over the RPC. So it should be easy to test. But we should make dastard commander aware of these changes before merging.

### DIFF
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -8,6 +8,8 @@ import (
 	"math"
 	"os"
 	"os/signal"
+	"os/user"
+	"path/filepath"
 	"sort"
 	"time"
 
@@ -95,7 +97,7 @@ func (ls *LanceroSource) Delete() {
 	}
 }
 
-// used to make sure the same device isn't used twice
+// contains is used to make sure the same device isn't used twice
 func contains(s []*LanceroDevice, e *LanceroDevice) bool {
 	for _, a := range s {
 		if a == e {
@@ -110,8 +112,6 @@ func contains(s []*LanceroDevice, e *LanceroDevice) bool {
 // be permanent, but I do think ClockMhz is necessarily the same for all cards.
 type LanceroSourceConfig struct {
 	FiberMask         uint32
-	ClockMhz          int
-	Nsamp             int
 	CardDelay         []int
 	ActiveCards       []int
 	AvailableCards    []int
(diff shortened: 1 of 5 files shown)


lancero_source.go
```diff
@@ -368,6 +370,46 @@ func roundint(x float64) int {
 	return int(x + math.Copysign(0.5, x))
 }
 
+type cringeGlobals struct {
+	SETT             int `json:"SETT"`
```
joefowler: Change Golang variable to `Settle` or `SETTLE`.

lancero_source.go
```diff
@@ -368,6 +370,46 @@ func roundint(x float64) int {
...
+	SquenceLength    int `json:"seqln"`
```
joefowler: Don't you mean `SequenceLength` (properly spelled)?

lancero_source.go
```diff
@@ -368,6 +370,46 @@ func roundint(x float64) int {
...
+	// jsonFile's content into 'cg'
+	err = json.Unmarshal(byteValue, &cg)
+	if err != nil {
+		return cringeGlobals{}, err
+	}
+	return cg, nil
+}
+
+// InjestCringeGlobals changes NSAMP, nrows and lsync to match the values from cg
+func (ls *LanceroSource) InjestCringeGlobals(cg cringeGlobals) {
```
joefowler: Spell `Ingest`. I'm not saying this in jest.

ggggggggg:
Oddly I think you commented on an earlier commit. When I went to make your changes `SETT` was named `Sett`, the `Injest...` method had been removed and `SequenceLength` was spelled correctly. Did you click on the "File Changed" tab?

joefowler:
I get confused in PR reviews. I must have reviewed the first commit, not the overall change. Hmm. Well, whatever. Looks fine to me.

PR #139, opened 2019-04-18
Roach data source is beginning to work

### LINKED ISSUE #127: Read Roach roach devices
We need to add support for roach devices. @joefowler 

Here are links to the implementation in matter: 
 * https://bitbucket.org/nist_microcal/nasa_daq/src/master/ndfb_server/RoachUDP.cpp
 * https://bitbucket.org/nist_microcal/nasa_daq/src/master/ndfb_server/RoachUDPProcessor.cpp

Do we have some software that generates Roach likes UDP packets, or any other documentation that would be useful? I'll email this issue to relevant people not on github.

joefowler:
We have a basically working version of Dastard that will handle ROACH packets. Fixes #127 .


joefowler:
Of course, we want to fix the fact that this version fails 100 tests.... Right?

ggggggggg:
I fixed the tests, which were related to signedness. There were two sources of truth about signedness, I deleted one of them. Now the only source of truth about signenedess is at creation of `DataSegment` deep inside a data source.

PR #141, opened 2019-04-29
Failure to ConfigureLanceroSource does not cause a panic

### LINKED ISSUE #140: Should not panic when cringeGlobals.json doesn't exist
I think it's a mistake to panic on an absent `cringeGlobals.json` file. It's fine to say that the Lancero source is unusable in this case, but I think we are just begging for confused questions from users if we panic.

I'll think about whether disabling Lancero sources is easy to do.... See PR #138 and the issue #133 for more information.

On my mac:

```
fowlerj@817jfmac:dastard$ go run cmd/dastard/dastard.go 
2019/04/29 09:17:39 cringeGlobalsPath /Users/fowlerj/.cringe/cringeGlobals.json
2019/04/29 09:17:39 Dastard is using config file /Users/fowlerj/.dastard/config.yaml
2019/04/29 09:17:39 ConfigureSimPulseSource: 2 chan, rate=6000.000
2019/04/29 09:17:39 Result is okay=true and state={2 chan, rate=6000.000}
2019/04/29 09:17:39 ConfigureTriangleSource: 1 chan, rate=10.000
2019/04/29 09:17:39 Result is okay=true and state={1 chan, rate=10.000}
2019/04/29 09:17:39 ConfigureLanceroSource: mask 0xffff  active cards: []
2019/04/29 09:17:39 Result is okay=false and state={0 MHz clock, 0 cards}
panic: open /Users/fowlerj/.cringe/cringeGlobals.json: no such file or directory

goroutine 1 [running]:
github.com/usnistgov/dastard.RunRPCServer(0x157c, 0x4707a01)
	/Users/fowlerj/go/src/github.com/usnistgov/dastard/rpc_server.go:600 +0x997
main.main()
	/Users/fowlerj/go/src/github.com/usnistgov/dastard/cmd/dastard/dastard.go:96 +0x1bd
E: 19-04-29 09:17:40 [8859]dangling 'PUB' socket created at :179
exit status 2
```

joefowler:
If `ConfigureLanceroSource` fails (e.g., because there's no cringe configuration file), then remember this error but don't panic. For a non-Lancero computer, this failure is normal behavior.

Instead of panicking, remember the error and refuse to start the source in the `Sample()` method. (Given that a non-Lancero machine won't have `/dev/lancero_user*` files, this will not actually come up in practice.)

Fixes #140.

### DIFF
--- a/lancero_source.go
+++ b/lancero_source.go
@@ -236,6 +236,10 @@ func (ls *LanceroSource) ConfigureMixFraction(mfo *MixFractionObject) ([]float64
 
 // Sample determines key data facts by sampling some initial data.
 func (ls *LanceroSource) Sample() error {
+	// Cannot start this process if configuration failed.
+	if ls.configError != nil {
+		return ls.configError
+	}
 	ls.dataBlockCount = 0
 	ls.nchan = 0
 	for _, device := range ls.active {
--- a/rpc_server.go
+++ b/rpc_server.go
@@ -126,6 +126,8 @@ func (s *SourceControl) ConfigureSimPulseSource(args *SimPulseSourceConfig, repl
 func (s *SourceControl) ConfigureLanceroSource(args *LanceroSourceConfig, reply *bool) error {
 	log.Printf("ConfigureLanceroSource: mask 0x%4.4x  active cards: %v\n", args.FiberMask, args.ActiveCards)
 	err := s.lancero.Configure(args)
+	// Remember any errors for later, when we try to start the source.
+	s.lancero.configError = err
 	s.clientUpdates <- ClientUpdate{"LANCERO", args}
 	*reply = (err == nil)
 	log.Printf("Result is okay=%t and state={%d MHz clock, %d cards}\n", *reply, s.lancero.clockMhz, s.lancero.ncards)
@@ -595,10 +597,8 @@ func RunRPCServer(portrpc int, block bool) {
 	var lsc LanceroSourceConfig
 	err = viper.UnmarshalKey("lancero", &lsc)
 	if err == nil {
-		err0 := sourceControl.ConfigureLanceroSource(&lsc, &okay)
-		if err0 != nil {
-			panic(err0)
-		}
+		sourceControl.ConfigureLanceroSource(&lsc, &okay)
+		// Don't panic on config errors: they are expected on a system w/o Lancero cards!
 	}
 	var asc AbacoSourceConfig
 	err = viper.UnmarshalKey("abaco", &asc)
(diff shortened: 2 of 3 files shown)


rpc_server.go
```diff
@@ -595,10 +597,8 @@ func RunRPCServer(portrpc int, block bool) {
 	var lsc LanceroSourceConfig
 	err = viper.UnmarshalKey("lancero", &lsc)
 	if err == nil {
-		err0 := sourceControl.ConfigureLanceroSource(&lsc, &okay)
-		if err0 != nil {
-			panic(err0)
-		}
+		sourceControl.ConfigureLanceroSource(&lsc, &okay)
+		// Don't panic on config errors: they are expected on a system w/o Lancero cards!
```
ggggggggg:
```suggestion
if err0 := nil {
    log.Println("Error during ConfigureLanceroSource:", err)
```

lancero_source.go
```diff
@@ -236,6 +236,10 @@ func (ls *LanceroSource) ConfigureMixFraction(mfo *MixFractionObject) ([]float64
 
 // Sample determines key data facts by sampling some initial data.
 func (ls *LanceroSource) Sample() error {
+	// Cannot start this process if configuration failed.
+	if ls.configError != nil {
```
ggggggggg: This check should go before the first line of `Start` not in `Sample`, the first line of `Start` called `SetStateStarting` which probably changes internal state we'd rather not change.

joefowler: Cannot set it in `Start`, because that's a function using only the methods of the interface `DataSource`. We would need to add a method to `DataSource` to accomplish this.

rpc_server.go
joefowler: Not sure what you're proposing here, but the if-test will never succeed. Are you trying to get a log message? I'd put that in the `ConfigureLanceroSource` if you want that.

lancero_source.go
joefowler: I suppose we could override `AnySource.SetStateStarting` method to a special version for `LanceroSource.SetStateStarting`. I'd rather keep it the way I had it at first: less confusing. But I do take your point about changing internal state.

ggggggggg: It's probably fine, the source isn't supposed to work anyway, maybe just add a comment about why it is done this way and how it "should" be done.

rpc_server.go
ggggggggg: Mostly I just don't want to have an unchecked error, we should check it, and do SOMETHING so I don't later come through and add a `panic` which is my default response to an unchecked error.

joefowler: Fair enough. I'll make it clearer. Probably the smart thing would be to not try to configure Lancero sources when no such device exists.

lancero_source.go
joefowler: Okay. See latest commit 2a59540. This is perhaps an improvement on the internal state in case errors are returned by `Sample` or `PrepareRun`.

PR #149, opened 2019-05-06
Refactoring of WritingState and AnySource.WriteControl

### LINKED ISSUE #144: Race conditions exist again
Try `go test -race` and you'll see them. They appear to arise in `(*SimPulseSource).ComputeWritingState()` and similarly for other source types.

joefowler:
This refactoring moves certain changes to the WritingState object into methods of that object, and uses a `sync.Mutex` to lock the object. This allows the dastard tests to pass without race conditions. Fixes #144 as far as I can tell.

Not only does it pass the tests, but when used with simulated pulse data, the LJH files and the experiment_state.txt files make sense.


writing_state.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// Stop will set the WritingState to be completely stopped
+func (ws *WritingState) Stop() error {
+	ws.Lock()
+	defer ws.Unlock()
+	ws.Active = false
+	ws.Paused = false
+	ws.FilenamePattern = ""
+	ws.setExperimentStateLabel(time.Now(), "STOP")
```
ggggggggg: check errors

joefowler: Good catch. Actually, tests fail if you check for and return any non-nil error returned by ws.setExperimentStateLabel in line 74. I did not figure out whether this was a mistake in the test, because it was easier to ignore the return value. But you're right, that's not the right approach. Let me see if the test is in error....

joefowler:
Okay, see whether commit 35d53330 addresses your concern. I think the behavior is better now. (I admit that I am confused about why the previous behavior passed our tests before today--it seems like it should have been a fail.)

PR #150, opened 2019-05-09
function arguments now *mat.Dense not mat.Dense

### LINKED ISSUE #136: Gonum mat.Dense is being used as a concrete value rather than a pointer
In a number of places you have functions taking parameters that are `mat.Dense`. The `mat.Dense` type should be passed as a pointer value. You are currently not doing anything that will break because of this, but you are opening up yourself to problems if people forget.

I was going to send a PR to fix this, but you have an interface that locks this in place, so I figured it would be better to let you deal with it.

ggggggggg:
fixes #136

### DIFF
--- a/process_data.go
+++ b/process_data.go
@@ -63,7 +63,7 @@ func (dsp *DataStreamProcessor) SetProjectorsBasis(projectors *mat.Dense, basis 
 
 // HasProjectors return true if projectors are loaded
 func (dsp *DataStreamProcessor) HasProjectors() bool {
-	return !dsp.projectors.IsZero()
+	return (dsp.projectors != nil) && (!dsp.projectors.IsZero())
 }
 
 // NewDataStreamProcessor creates and initializes a new DataStreamProcessor.
--- a/process_data.go
+++ b/process_data.go
@@ -19,12 +19,12 @@ type DataStreamProcessor struct {
 	LastTrigger          FrameIndex
 	LastEdgeMultiTrigger FrameIndex
 	stream               DataStream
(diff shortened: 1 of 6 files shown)


process_data.go
```diff
@@ -55,8 +55,8 @@ func (dsp *DataStreamProcessor) SetProjectorsBasis(projectors mat.Dense, basis m
 	if brows != dsp.NSamples {
 		return fmt.Errorf("basis has wrong size, has rows: %v, want: %v", brows, dsp.NSamples)
 	}
-	dsp.projectors = projectors
-	dsp.basis = basis
+	dsp.projectors = *projectors
+	dsp.basis = *basis
```
kortschak: Make `dsp.basis` and `dsp.projectors` hold `*mat.Dense` and these derefs will go away.

ggggggggg: that makes sense. I didn't do that because I later use `IsZero` and I couldn't figure out how to initialize them without calling `NewDense` which requires meaningless (since I don't have it yet) size information?

ggggggggg: I guess I could do `var temp mat.Dense;dsp.projectors=&temp`?

kortschak: Yes, that's how I would do it.

kortschak:
I haven't looked closely at the code, but you may want to make [this line](https://github.com/usnistgov/dastard/blob/cd9cb53f620112f9c6a9f60be0c7890e3402c885/process_data.go#L66) say `return dsp != nil && !dsp.projectors.IsZero()` to avoid the possibility of a nil pointer deref.

PR #157, opened 2019-06-14
position info to file headers

### LINKED ISSUE #108: Propagate more info to the LJH + off headers
An incomplete list: 

1. the filename the projectors come from
2. the pixel location

ggggggggg:
Pass position info to `LJH22` and `off` file headers. Having `MapServer` separate from `SourceControl` source control doesn't seem helpful, as I have to keep a reference to `MapServer` to be able to know anything about pixels. I didn't change it because I didn't want to change the tests, but I think we should just accept that `SourceControl` is a god object.

fixes #108

### DIFF
--- a/data_source.go
+++ b/data_source.go
@@ -502,6 +502,15 @@ func (ds *AnySource) writeControlStart(config *WriteControlConfig) error {
 		rowNum := rccode.row()
 		colNum := rccode.col()
 		fps := 1
+		var pixel Pixel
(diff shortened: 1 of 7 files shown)


ggggggggg:
My comment was meant to say that I think the functionality of mapserver should have just been built into source control.

I don't know about the units. I originally added a units field in the off and ljh headers, but there is no unit info to provide. We could add a units field to the map files and pass that through.

Issue #163, opened 2019-08-08
Dastard shouldn't crash on missing data


ggggggggg:
Currently, if dastard notices missing data (eg the frame bits are misaligned in lancero data), it panics. We should try hard to avoid this case, but we shouldn't panic. Instead we should use the CPU clock to estimate how much data is missing, update the framecount accordingly, and note the missing data in a file.

Ideally this would be implemented in a way that works for any data source so it doesn't need to be reimplemented over and over.

joefowler:
Fixed by #166

PR #166, opened 2019-08-29
Handle dropped data

ggggggggg:
When the first frame bit  in lancero source is in the incorrect location, we note a data drop. Then we estimate how many frames were dropped based on cpu time and sample rate. Then we write to a file ending in "data_drop.txt" with the `framenum` and `estimated number of frames dropped`. We also send a DATADROP message and flush the data drop file every 10 seconds. I modeled the code on how external triggers are handled to the extent possible.

Data drops for multiple device lancero sources panic for now, but it should be easy to handle.


joefowler:
Fixes #163

PR #170, opened 2019-09-06
Map info to ljh3

ggggggggg:


ggggggggg:
I was seeing an occasionaly "heatbeat missing" with simpulsesource after long periods (like overnight). I thought it could be a bug with the timing code, and tickers are easier to reason about.

ggggggggg:
It could also be a bug in dastard commander.

joefowler:
Would you be willing to merge the #168 (Abaco PR) before this one? Seems easier to do large than small.

ggggggggg:
sure ill review it then merge it

PR #175, opened 2020-01-06
write pretrig delta in off files, increment version

ggggggggg:
As discussed I added the pretrig delta to the off files. This should allow the equivalent of the pretrig rms cut in a simple projector independent way.

### DIFF
--- a/process_data.go
+++ b/process_data.go
@@ -195,11 +195,18 @@ func (dsp *DataStreamProcessor) AnalyzeData(records []*DataRecord) {
 			}
 		}
 
-		var val float64
+		var val float64        // used to calculate pretrigger mean, then reused in the next loop
+		var valPTDelta float64 // used to calculate pretrigger delta
+		// slope = dot(x,y.-y[0])/z where z = dot(x,x) and x = [0, 1, 2, ..., N-1]/(N-1), where .-y[0] is elementwise subtraction of the first element
+		np := rec.presamples
+		a := float64(np*(np-1)*(2*np-1)) / float64(6) / float64((np - 1)) // a is the sum of np^2 from 0 to np-1 over (np-1)
+		// a converts valPTDelta to the change over the length of the pretrigger
 		for i := 0; i < rec.presamples; i++ {
 			val += dataVec.AtVec(i)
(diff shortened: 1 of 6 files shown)


process_data.go
```diff
@@ -195,11 +195,18 @@ func (dsp *DataStreamProcessor) AnalyzeData(records []*DataRecord) {
...
+		// slope = dot(x,y.-y[0])/z where z = dot(x,x) and x = [0, 1, 2, ..., N-1]/(N-1), where .-y[0] is elementwise subtraction of the first element
+		np := rec.presamples
+		a := float64(np*(np-1)*(2*np-1)) / float64(6) / float64((np - 1)) // a is the sum of np^2 from 0 to np-1 over (np-1)
+		// a converts valPTDelta to the change over the length of the pretrigger
 		for i := 0; i < rec.presamples; i++ {
 			val += dataVec.AtVec(i)
+			valPTDelta += (dataVec.AtVec(i) - dataVec.AtVec(0)) * float64(i)
 		}
 		ptm := val / float64(rec.presamples)
+		rec.pretrigDelta = valPTDelta / a
```
joefowler: I don't think this computation is right. Let me work it out and see if I can show this by further testing.

process_data.go
```diff
@@ -198,15 +198,19 @@ func (dsp *DataStreamProcessor) AnalyzeData(records []*DataRecord) {
...
+		for i := 0; i < npre; i++ {
 			val += dataVec.AtVec(i)
-			valPTDelta += (dataVec.AtVec(i) - dataVec.AtVec(0)) * float64(i)
+			valPTDelta += (dataVec.AtVec(i) - d0) * (float64(i) - xmean)
 		}
-		ptm := val / float64(rec.presamples)
-		rec.pretrigDelta = valPTDelta / a
+		ptm := val / float64(npre)
+		rec.pretrigDelta = valPTDelta * 12.0 / float64(npre*(npre+1))
+		if npre <= 1 {
```
ggggggggg: The other assignment should be an else clause here

joefowler: Okay. See next commit

ggggggggg:
Thanks. I think merge as soon as tests pass

PR #187, opened 2020-04-20
Update install instructions and the make install target

joefowler:
1. Updated the installation instructions on the READMe.
1. Improved the `make install` so it installs the binary with our version tag and git hash built in.
1. Simplified (and sped up) the Travis CI testing now that we are using a pure-go ZMQ implementation. Because of this, we don't need to install the obscure non-go dependencies of czmq and libzmq3 and libsodium.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,3 @@ go:
   - 1.14.x
 
 go_import_path: github.com/usnistgov/dastard
-
-addons:
-  apt:
-    sources:
-    - sourceline: 'deb http://download.opensuse.org/repositories/network:/messaging:/zeromq:/git-draft/xUbuntu_16.04/ ./'
-      key_url: 'http://download.opensuse.org/repositories/network:/messaging:/zeromq:/git-draft/xUbuntu_16.04/Release.key'
-    packages:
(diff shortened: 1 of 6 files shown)


ggggggggg:
Can we just have `make install` copy to `/usr/local/bin` so we dont need to modify `.bashrc`?

joefowler:
1. You could, but `/usr/local/bin` isn't in everyone's path on every Unix system either, so you still have to at least have users _check_ on that in the install instructions.
1. It's not the go-ish place to put it.
1. It would require a sudo step.

ggggggggg:
Lets add a line like this so they dont need an editor: `echo "source ~/qsp/bin/activate" >> ~/.bashrc`?

joefowler:
You could, but this sort of trick can lead to an ugly `.bashrc` if one runs it more than once. You could do a check before adding the line, but that might be tricky to get right.

I think having users do a one-time step, even one that requires an editor, is preferred to an automated "solution" that could fail in a bunch of ways that are hard to predict.

ggggggggg:
Just add it as an optional or something. I'm the one who does this setup most of the time, and I want it to be easy. Also, people who care if their  `.bashrc` looks nice know how to fix it and can choose to use an editor.

Issue #192, opened 2020-05-11
repeated zmq errors 


ggggggggg:
Looking at the output of dastard over a few days I see many (100?) messages like the following 

```zmq4: could not open a ZMTP connection with "tcp://*:5502": zmq4: could not initialize ZMTP connection:
    github.com/go-zeromq/zmq4.Open
        /home/pcuser/go/pkg/mod/github.com/go-zeromq/zmq4@v0.9.0/conn.go:92
  - zmq4: could not exchange greetings:
    github.com/go-zeromq/zmq4.(*Conn).init
        /home/pcuser/go/pkg/mod/github.com/go-zeromq/zmq4@v0.9.0/conn.go:104
  - zmq4: could not recv greeting:
    github.com/go-zeromq/zmq4.(*Conn).greet
        /home/pcuser/go/pkg/mod/github.com/go-zeromq/zmq4@v0.9.0/conn.go:146
  - could not read ZMTP greeting:
    github.com/go-zeromq/zmq4.(*greeting).read
        /home/pcuser/go/pkg/mod/github.com/go-zeromq/zmq4@v0.9.0/protocol.go:98
  - read tcp 132.163.82.105:5502->132.163.5.35:41178: read: connection reset by peer
```

@sbinet What does this mean? Does this correspond to a line in dastard (if so, how can I find out which one) or is this an internal message from `go-zeromq`?

sbinet:
This is the display of an error originating from zmq (but I suspect it's being displayed by dastard).
This error happens when opening a new zmq connection and the peer dropped off during the zmq greeting handshake.

ggggggggg:
Thanks. I suspect this is something the previous zmq library just didn't show us, and it is due to a security scanner trying to connect in some way. Is there a way to suppress the output?

sbinet:
there are 2 places where `dastard` is creating `zmq4.Socket`s:
- https://github.com/usnistgov/dastard/blob/9b4ceeecc4d331502713b7007e24833e57da5b3c/client_updater.go#L49..L52
- https://github.com/usnistgov/dastard/blob/9b4ceeecc4d331502713b7007e24833e57da5b3c/publish_data.go#L378..L385

if it's not coming from there, then it's indeed coming from zmq4, here:
- https://github.com/go-zeromq/zmq4/blob/de8318e9c31dc075627022e399766a39e0b5e2aa/socket.go#L201

then, a simple "fix" would be to create a `/dev/null` logger:
```go
[...]

PR #193, opened 2020-05-26
Back to czmq

ggggggggg:
I found that when using the go zeromq implementaiton many data packets were dropped mysteriously, so I switched back to the  czmq implementation.

I also found an unrelated bug with some duplicate code in `lancero.go` that I removed.


joefowler:
Fixes #192.

.travis.yml
```diff
@@ -8,3 +8,11 @@ go:
...
 
 go_import_path: github.com/usnistgov/dastard
+
+addons:
+  apt:
+    sources:
+    - sourceline: 'deb http://download.opensuse.org/repositories/network:/messaging:/zeromq:/git-draft/xUbuntu_16.04/ ./'
+      key_url: 'http://download.opensuse.org/repositories/network:/messaging:/zeromq:/git-draft/xUbuntu_16.04/Release.key'
+    packages:
+    - libczmq-dev libzmq3-dev libsodium-dev
```
joefowler, line 18: Add a linefeed to last line.

joefowler:
Do you want me to investigate #194 before we merge this PR? I am starting to do that.

ggggggggg:
I pushed a commit with the code duplication removed, I'll then re-ran autotune on horton and it works. I guess I had just removed the code duplication locally and forgot to commit it. Good catch.

ggggggggg:
No, #194 cant be separate. It doesn't obviously break anything.

PR #195, opened 2020-05-26
Fix194

### LINKED ISSUE #194: CURRENTIME message empty sometimes
dcom is printing this out fairly regularly: `Error processing status message [topic,msg]: CURRENTTIME, 
`

Which I think means the `msg` component of CURRENTIME is empty.

joefowler:
There is a problem: I cannot figure out a simple way to test this fix and prevent regression. Right now we are not mocking up a status message receiver, which would be needed to test this fix. I think that mock-up would not be worth the trouble.

Fixes #194. (You'll have to take my word for it!)

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -3,8 +3,9 @@
 **0.2.7** in progress in April 2020
 * Use package `zmq4`, a pure Go implementation of ZMQ.
 * Simplified testing on Travis because of that.
-* Then revert to `gozmq` because `zmq4` was dropping packets.
+* Then revert to `gozmq` because `zmq4` was dropping packets. (Fixes 192)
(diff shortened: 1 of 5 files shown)


ggggggggg:
the main fix is to just not send the CURRENTTIME message to clients?

joefowler:
Okay smart guy. Yes, that's the fix.

You see, strings containing spaces seem to be allowed in YAML files but have to be quoted in JSON. The latter is needed for sending to clients, at least to the Python client. As far as I can tell, CURRENTTIME is the only message we were sending that put this rule to the test. Now, I could have figured out a way to make the necessary transformation (adding quotes before sending), but that seemed like a waste of effort, as we simply drop that message at the client end.

ggggggggg:
Makes sense, makes the clients simpler.

PR #203, opened 2020-08-19
Handle Abaco channel groups as mostly independent experiments.

### LINKED ISSUE #190: Find sample rate with Abaco data reader splitting channels into 2+ packets
When the Abaco is splitting the channels across 2 or more packets, we are not finding the data rate correctly. Fix this, possibly by testing only the first group for rate info.

### LINKED ISSUE #202: AbacoSource needs to fill in data where packets are dropped
Firmware team insists that dropping of packets is theoretically possible and should be handled. Proposal at the moment is to replace missing packets with packets containing constant data.

joefowler:
Reworks how the `AbacoSource` handles "devices" (now, `AbacoRing` objects) and separates that from how it handles the channel groups (`AbacoGroup` objects). Packets from the former are distributed to the appropriate queue of the latter.

- Handles missing Abaco packets (by filling in dummy values).
- Achieves crude (~ms-scale?) synchronization between channel groups.
- Allows for channel numbering with (potentially large) gaps between numbers in successive groups.
- Fixes #190.
- Fixes #202.

### DIFF
--- a/abaco.go
+++ b/abaco.go
@@ -49,7 +49,7 @@ type AbacoGroup struct {
 	unwrap     []*PhaseUnwrapper
 	queue      []*packets.Packet
 	lasttime   time.Time
-	seqnumsync uint32 // Ideally, all AbacoGroups are in sync at this packet sequence number.
+	seqnumsync uint32 // Global sequence number is referenced to this group's seq number at seqnumsync
 }
 
 // NewAbacoGroup creates an AbacoGroup given the specified GroupIndex.
@@ -88,6 +88,9 @@ func (group *AbacoGroup) samplePackets() error {
 				tsInit.T = ts.T
 				tsInit.Rate = ts.Rate
 				snInit = p.SequenceNumber()
+				// For now: sync by assuming first packet seen by each group is simultaneous
+				// TODO: eventually want to sync to the _sample_ level, not the packet level.
+				group.seqnumsync = snInit
 			}
 			if tsFinal.T < ts.T {
 				tsFinal.T = ts.T
@@ -120,11 +123,20 @@ func (group *AbacoGroup) samplePackets() error {
 	return nil
 }
 
+// firstSeqNum returns the global sequence number for the first packet in the group's packet queue,
+// meaning the sequence number referenced to the group.seqnumsync offset.
+func (group *AbacoGroup) firstSeqNum() (uint32, error) {
+	if len(group.queue) == 0 {
+		return 0, fmt.Errorf("No packets in queue")
+	}
+	return group.queue[0].SequenceNumber() - group.seqnumsync, nil
+}
+
 // fillMissingPackets looks for holes in the sequence numbers in group.queue, and replaces them with
 // pretend packets.
-func (group *AbacoGroup) fillMissingPackets() uint32 {
+func (group *AbacoGroup) fillMissingPackets() (numberAdded int) {
 	if len(group.queue) <= 1 {
-		return 0
+		return
 	}
 	sn0 := group.queue[0].SequenceNumber()
 	for i := 1; i < len(group.queue); i++ {
@@ -133,14 +145,16 @@ func (group *AbacoGroup) fillMissingPackets() uint32 {
 		if sn-sn0 != uint32(i) {
 			fmt.Printf("Warning! Need new packet here")
 			panic("fillMissingPackets not implemented yet")
+			numberAdded++
 			// TODO: put in an extra packet here.
 		}
 	}
 	// TODO: find and fill holes even if they are between past history and the first packet
-	return sn0 - group.seqnumsync
+	return
 }
 
 // trimPacketsBefore removes leading packets from the queue that "predate" firstSn
+// firstSn is a "global sequence number", thus relative to the group.seqnumsync
 func (group *AbacoGroup) trimPacketsBefore(firstSn uint32) {
 	firstSn += group.seqnumsync
 	for ;; {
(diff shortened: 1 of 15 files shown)


client_updater.go
```diff
@@ -26,22 +26,21 @@ type ClientUpdate struct {
...
+	"CURRENTTIME": {},
 }
 
 // nologMessages is a set of message names that you don't log to the terminal, because they
 // are too long or too frequent to bother with.
 var nologMessages = map[string]struct{}{
-	"TRIGGERRATE":    {},
+	"TRIGGERRATE":     {},
 	"CHANNELNAMES":    {},
-	"ALIVE":    {},
```
ggggggggg, line 37: why is all this formatting changing? did gofmt rules change or something?

abaco.go
```diff
@@ -15,59 +16,290 @@ import (
...
-// ByGroup implements sort.Interface for ChannelGroup slices
-type ByGroup []ChannelGroup
-func (g ByGroup) Len() int { return len(g) }
-func (g ByGroup) Swap(i, j int) { g[i], g[j] = g[j], g[i] }
+// ByGroup implements sort.Interface for []GroupIndex so we can sort such slices.
+type ByGroup []GroupIndex
+
+func (g ByGroup) Len() int           { return len(g) }
+func (g ByGroup) Swap(i, j int)      { g[i], g[j] = g[j], g[i] }
 func (g ByGroup) Less(i, j int) bool { return g[i].firstchan < g[j].firstchan }
```
ggggggggg, line 39: What if one of them has `firstchan = 1` and `nchan=10000000` and the other has `firstchan = 2`. Then one is not strictly less than the other.

abaco.go
```diff
@@ -86,138 +318,79 @@ func (device *AbacoDevice) ReadAllPackets() ([]*packets.Packet, error) {
...
-	maxDelay := time.Duration(2000 * time.Millisecond)
-	timeOut := time.NewTimer(maxDelay)
-
-	// Capture timestamp and sample # for a range of packets. Use to find rate.
-	var tsInit, tsFinal packets.PacketTimestamp
-	var snInit, snFinal uint32
-	samplesInPackets := 0
+	// Now get the data we actually want: fresh data. Run for at least
+	// a minimum time or a minimum number of packets.
+	const minPacketsToRead = 100 // Not sure this is a good minimum?
```
ggggggggg, line 343: what is needed for success? one from each channel group? would be nice to have a comment describing what is neccesary to learn all you need

abaco.go
```diff
@@ -86,138 +318,79 @@ func (device *AbacoDevice) ReadAllPackets() ([]*packets.Packet, error) {
...
+//------------------------------------------------------------------------------------------------
+
+// AbacoSource represents all AbacoRing ring buffers that can potentially supply data, as well
+// as all AbacoGroups that are discovered in the SampleData phase.
 type AbacoSource struct {
-	devices     map[int]*AbacoDevice
-	Ndevices    int
-	active      []*AbacoDevice
+	Nrings int // number of available ring buffers
+	arings map[int]*AbacoRing
```
ggggggggg, line 388: are they not guaranteed to be sequentially ordered? I guess if we had four and chose to read 1 and 3 then they would not be sequentially numbered.

abaco.go
```diff
@@ -304,37 +478,92 @@ func (as *AbacoSource) Configure(config *AbacoSourceConfig) (err error) {
...
+				as.groups[cidx] = NewAbacoGroup(cidx)
+				as.nchan += cidx.nchan
+			}
 		}
+		as.distributePackets(results.allpackets, now)
 	}
-	for _, device := range as.active {
-		as.nchan += device.nchan
+
+	// Verify that no channel # appears in 2 groups.
```
ggggggggg, line 527: this addresses my worries about the sorting I guess

ggggggggg:
Looks like travis is telling us the tests ran again, did you do something?

client_updater.go
joefowler: I think my IDE (Atom) was supposed to be running go fmt and hasn't been for a long time. I tried running it at the command line, which yielded results on only some files. No idea why. Not happy about it, but not willing to track down the problem.

abaco.go
joefowler: Right. This is _an_ ordering rule, not _the only conceivable_ ordering rule.

Later, AbacoGroup objects will be sorted in this way, but also checked to ensure that no channel appears in more than one group. Given that check, the rule here is actually the only sensible one.

joefowler:
Travis ran 2 tests: one on the push to channel_groups branch, one to the hypothetical result of the PR. Nothing beyond those, I think.

abaco.go
joefowler: Good point. I will attempt to spell it out in a comment, though I'm not sure I really know what's needed. We need "some data", with at least 2 packets. That's because we need the time between packets (as time-stamped by the firmware on the Abaco) to estimate sample rates.

joefowler:
I agree that we need to count and maybe log data drops. I decided that ought to be a new issue, so see #204. Add to that if you like.

ggggggggg:
LGTM. I thought the sample rate was the sort of thing that would be in an abaco packet header?

ggggggggg:
Also maybe the number of channel groups?

joefowler:
Nope. Apparently these (sample rate, # of groups) are far easier for us to glean in software than for the firmware to know and then broadcast. 

Sample rate is hard to know because it relies on the behavior of an external source (the flux ramp). I think there are limits to how fast it can be, but basically the Abaco cannot change the sample rate. And it can only know what the rate was in the past, not in the future. If it computes the rate, it has no way to know that it will be the same in the future. Anyway, that's my understanding.

The number of groups is also hard to know, simply because no group knows about the others. There are firmware blocks that know this information, but these blocks don't generate the packets!  They could generate their own, different kind of packets, but that's now how it works now.

I think M said that if we are already designing Dastard with this concept of "sample for a while, learn some stuff, then actually take data" then we might as well use it to allow simpler firmware designs.

PR #205, opened 2020-08-26
use EdgeMultiLeve instead of EdgeLevel and and EdgeMultiDisableZeroTh…

ggggggggg:
add the ability to turn off the zero threshold trigger, and have edge-multi use its own theshold value


ggggggggg:
also improve mac os install instructions

PR #209, opened 2020-09-02
Handle dropped data correctly, with logging

### LINKED ISSUE #204: AbacoSource or AbacoGroup needs to log dropped packets
See #203 comments. We should log somewhere that a packet was dropped and filled in. Lancero source already does this. We could consider whether a more thorough, formal system for logging problems is called for.

joefowler:
Bahama now can drop a specified percentage of data packets. Dastard handles correctly, with logging to `~/.dastard/logs/problems.log` and log-rotation provided by the [lumberjack](https://github.com/natefinch/lumberjack) package.  Fixes #204.

Also ran `go fmt` for the first time in a while.

### DIFF
--- a/abaco.go
+++ b/abaco.go
@@ -146,8 +146,7 @@ func (group *AbacoGroup) fillMissingPackets() (numberAdded int) {
 	for _, p := range group.queue {
 		sn := p.SequenceNumber()
 		for snexpect < sn {
-			val := p.ReadValue(0) // Fill data with first sample from next non-missing packet
-			pfake := p.MakePretendPacket(snexpect, val)
+			pfake := p.MakePretendPacket(snexpect, group.nchan)
 			newq = append(newq, pfake)
 			numberAdded++
 			snexpect++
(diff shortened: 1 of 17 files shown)


ggggggggg:
This still doesn't set the "dropped frames" value in `DataSegment` or otherwise communicate that there was a dropped frame up to the point where it can be logged. https://github.com/usnistgov/dastard/blob/c86f1a51deba5406a16d20a15e0067e36aeb9e91/lancero_source.go#L803


I'm guessing you mean something else by "handle dropped data correctly"?

joefowler:
Oops. I forgot to also log dropped Lancero frames. I was focused on Abaco. I'll get back to Lancero...

ggggggggg:
Ohh I see you implemented a seperate problem logger. I don't like that solution, I think it should go into the same directory as the ljh and off files so its part of the dataset. That how I made the lancero one work.

joefowler:
Okay, I will figure out how to use the `AnySource.HandleDataDrop` mechanism to log Abaco data drops in a fashion similar to Lancero ones.

ggggggggg (approved):

PR #215, opened 2020-09-18
Channel groups

joefowler:
Handle channel groups instead of only consecutive channel numbering.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,6 +1,9 @@
 ## DASTARD Versions
 
-**0.2.7** in progress in April 2020
+**0.2.8** Started September 2020
+* Remove idea of rows/columns from Dastard (use only in `LanceroSource`). Use chan groups otherwise.
+
+**0.2.7** September 10, 2020
 * Use package `zmq4`, a pure Go implementation of ZMQ.
 * Simplified testing on Travis because of that.
 * Then revert to `gozmq` because `zmq4` was dropping messages (issue 192).
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,8 +1,9 @@
 ## DASTARD Versions
 
 **0.2.8** Started September 2020
-* Remove idea of rows/columns from Dastard (use only in `LanceroSource`). Use chan groups otherwise (issue 214).
+* Remove idea of rows/columns from Dastard (used only in `LanceroSource`). Use chan groups otherwise (issue 214).
 * Lancero source can have more flexible channel numbers, such as 12043 for card 1, column 2, row 43 (issue 212).
+* Testing of the flexible channel numbering.
 
 **0.2.7** September 10, 2020
 * Use package `zmq4`, a pure Go implementation of ZMQ.
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,7 +1,8 @@
 ## DASTARD Versions
 
 **0.2.8** Started September 2020
-* Remove idea of rows/columns from Dastard (use only in `LanceroSource`). Use chan groups otherwise.
+* Remove idea of rows/columns from Dastard (use only in `LanceroSource`). Use chan groups otherwise (issue 214).
+* Lancero source can have more flexible channel numbers, such as 12043 for card 1, column 2, row 43 (issue 212).
 
 **0.2.7** September 10, 2020
 * Use package `zmq4`, a pure Go implementation of ZMQ.
--- a/abaco.go
+++ b/abaco.go
@@ -532,23 +532,6 @@ func (as *AbacoSource) Sample() error {
 	sort.Sort(ByGroup(keys))
 	as.groupKeysSorted = keys
 
-	// Fill the channel names and numbers slices
-	// For rowColCodes, treat each channel group as a "column" and number chan within it
-	// as rows 0...g.nchan-1.
-	as.chanNames = make([]string, 0, as.nchan)
-	as.chanNumbers = make([]int, 0, as.nchan)
-	as.rowColCodes = make([]RowColCode, 0, as.nchan)
-	ncol := len(keys)
-	for col, g := range as.groupKeysSorted {
-		for row := 0; row<g.Nchan; row++ {
-			cnum := row + g.Firstchan
-			name := fmt.Sprintf("chan%d", cnum)
-			as.chanNames = append(as.chanNames, name)
-			as.chanNumbers = append(as.chanNumbers, cnum)
-			as.rowColCodes = append(as.rowColCodes, rcCode(row, col, g.Nchan, ncol))
-		}
-	}
-
 	// Each AbacoGroup should process its sampled packets.
 	for _, group := range as.groups {
 		group.samplePackets()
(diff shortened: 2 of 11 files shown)


VERSIONS.md
```diff
@@ -1,6 +1,10 @@
 ## DASTARD Versions
 
-**0.2.7** in progress in April 2020
+**0.2.8** Started September 2020
+* Remove idea of rows/columns from Dastard (use only in `LanceroSource`). Use chan groups otherwise (issue 214).
```
ggggggggg, line 4: use should be used
otherwise seems to imply you still use row/columns sometimes.... maybe Replace with chan groups which are more general.

abaco.go
```diff
@@ -556,12 +540,12 @@ func (as *AbacoSource) Sample() error {
...
 	ncol := len(keys)
 	for col, g := range as.groupKeysSorted {
-		for row := 0; row<g.nchan; row++ {
-			cnum := row + g.firstchan
+		for row := 0; row<g.Nchan; row++ {
+			cnum := row + g.Firstchan
 			name := fmt.Sprintf("chan%d", cnum)
 			as.chanNames = append(as.chanNames, name)
 			as.chanNumbers = append(as.chanNumbers, cnum)
-			as.rowColCodes = append(as.rowColCodes, rcCode(row, col, g.nchan, ncol))
```
ggggggggg, line 564: rowColCodes always seemed a bit over engineered to me, but I guess they work fine so thats not super relevant. But should they be renamed to "channelGroupCodes" or something?

data_source.go
```diff
@@ -718,11 +742,20 @@ func (ds *AnySource) PrepareRun(Npresamples int, Nsamples int) error {
 	if ds.nchan <= 0 {
 		return fmt.Errorf("PrepareRun could not run with %d channels (expect > 0)", ds.nchan)
 	}
+
+	// Set up things that specific sources set up only if they have unusual settings.
 	if ds.channelsPerPixel < 1 {
```
ggggggggg, line 747: What case does this handle? This seems like a silent error handling case, maybe we should remove it and be explicit in all cases that it handles?

lancero_source.go
```diff
@@ -121,6 +124,9 @@ type LanceroSourceConfig struct {
 	CardDelay         []int
 	ActiveCards       []int
 	ShouldAutoRestart bool
+	FirstRow          int // Channel number of the 1st row (default 1)
+	ChanSepCards      int // Channel separation between cards (or 0 to indicate number sequentially)
+	ChanSepColumns    int // Channel separation between columns (or 0 to indicate number sequentially)
```
ggggggggg, line 129: If I want your example of `12034` for card 1, col 2, row 34, I put in FirstRow=1, ChanSepCards = 10000, ChanSepColumns=1000?

lancero_source.go
```diff
@@ -282,26 +291,58 @@ func (ls *LanceroSource) Sample() error {
...
-			ls.rowColCodes[i+j] = rcCode(row, col, device.nrows, device.ncols)
-			ls.rowColCodes[i+j+1] = ls.rowColCodes[i+j]
+	// Check that ls.chanSepColumns and chanSepCards are appropriate,
+	// i.e. large enough to avoid channel number collisions.
+	if ls.chanSepColumns > 0 {
+		for _, device := range ls.active {
+			if device.nrows > ls.chanSepColumns {
+				log.Printf("Warning: /dev/lancero%d has %d rows, which exceeds ChanSepColumns (%d). Setting latter to 0.",
+							device.devnum, device.nrows, ls.chanSepColumns)
+				ls.chanSepColumns = 0
```
ggggggggg, line 301: This is silent error handling, return an error instead.

lancero_source.go
```diff
@@ -282,26 +291,58 @@ func (ls *LanceroSource) Sample() error {
...
+	if ls.chanSepCards > 0 {
+		for _, device := range ls.active {
+			colsep := device.nrows
+			if ls.chanSepColumns > 0 {
+				colsep = ls.chanSepColumns
+			}
+			if device.nrows*device.ncols > ls.chanSepCards {
+				log.Printf("Warning: /dev/lancero%d needs %d channels, which exceeds ChanSepCards (%d). Setting latter to 0.",
+							device.devnum, colsep*device.ncols, ls.chanSepCards)
+				ls.chanSepCards = 0
```
ggggggggg, line 315: Same here, return an error.

lancero_source.go
ggggggggg: Also it seems like we want an assert is 0 or positive?

lancero_source.go
```diff
@@ -282,26 +291,58 @@ func (ls *LanceroSource) Sample() error {
...
-	for i := 1; i < ls.nchan; i += 2 {
-		cnum := 1+i/2
-		ls.chanNames[i-1] = fmt.Sprintf("err%d", cnum)
-		ls.chanNames[i] = fmt.Sprintf("chan%d", cnum)
-		ls.chanNumbers[i-1] = cnum
-		ls.chanNumbers[i] = cnum
+	index := 0
+	cnum := ls.firstRowChanNum
+	for _, device := range ls.active {
+		if ls.chanSepCards > 0 {
```
ggggggggg, line 327: This conditional should not be required, if it is zero, it's fine. You should have already handled the negative and other error cases.  Maybe all the channel numbering logic should be factored out into a function that can easily be tested?

lancero_source.go
```diff
@@ -282,26 +291,58 @@ func (ls *LanceroSource) Sample() error {
...
+			cnum = device.devnum*ls.chanSepCards + ls.firstRowChanNum
+		}
+		for col := 0; col < device.ncols; col++ {
+			if ls.chanSepColumns > 0 {
```
ggggggggg, line 331: same with this.

ggggggggg (changes requested):
Two main comments.
1. There is some silent error handling, just return errors and let them bubble up back through the RPC so the user knows something was wrong right away.
2. The channel numbering code is long enough it could be factored out into it's own function which could then have a few basic tests. This is less important than 1.

VERSIONS.md
joefowler: Well, I do use it still, and the only reason is that it goes into the headers of LJH and OFF files. I wish it didn't. Is there some way we could omit that or something? Does MASS really really need that?

If we didn't need the file to know the row/col numbers, then we could simplify a lot. But I fear that we really do need them or want them in the file. How we'll do the analogue in LJH/OFF files for Abaco is another thing that I don't know.

Maybe we can generalize that header, so it basically says "this is chan 4, and it's in a channel group with channels from 1-8". But we are treating files as if they know nothing about array size/geometry. Maybe we should do a phone call?

lancero_source.go
joefowler: Right. Whereas if you were handling only 32 rows at a time, and this is row 2 in the second group. In that case, you'd say FirstRow=33, but the other two would be like you said.

lancero_source.go
joefowler: It _is_ required. If it's zero, we should not be resetting the cnum counter; we should just be counting sequentially without gaps. Removing the test would reset the counter to the same value for each card.

lancero_source.go
joefowler: Same as above--I dispute your assertion. I do see the point that factoring this out into a new method would make it clearer and testable.

data_source.go
joefowler: Lines 746-757 handle the fact that many data source concrete classes do not currently
* Set `channelsPerPixel=1`.
* Set channel names and numbering arrays.
* Set up the `ds.groupKeysSorted` map.

So we have two choices:
1. Copy all of this code into all the concrete classes.
2. Put this here, to catch concrete classes that didn't set non-default values.

I feel like it would be nice to have the default values be set up earlier in the process, and merely overwritten by setup of concrete classes. The problem with that is that this default setup can't run until we "sample" the source and learn the number of channels. Anyway, this is my compromise.

joefowler: To be clear: only Lancero needs to set `channelsPerPixel=2`. Only Lancero has "Chan X" and "Err X" channel names. Only Lancero and Abaco have complicated channel groups allowed (the others would have the trivial case: a single from from 1 to N).

joefowler: Actually, I have a better idea. If there's an `AnySource.setUpChannelGroupingNaming` or something, then we could put all of this stuff in it, and the LanceroSource and AbacoSource could override it. I'll do that. This might fit well with the factoring-out-and-testing idea.

joefowler: Just pushed changes that add an `AnySource.PrepareChannels()` and methods of `LanceroSource` and `AbacoSource` that override it. Also added tests for the TDM version of the logic.

lancero_source.go
joefowler: Errors now returned here and in the next comment. Also added an error if either `chanSepCards` or Columns is negative.

abaco.go
joefowler: Decided on the phone to leave these as-is for now. Hope to change later, when we have more experience with Abaco and channel groups.

VERSIONS.md
joefowler: Agreed to continue using row/columns...for now.

lancero_source.go
joefowler: With testing, I hope it's more clear the logic is correct, even if it's basically unreadable!

ggggggggg (approved):

PR #221, opened 2020-10-24
Speed up Abaco data processing.

### LINKED ISSUE #218: Abaco source, when stopped, won't restart
Both M and I observe that if you stop the Abaco source (generated by Bahama), then you get an error if you try to restart. I don't know how universal this is. Somehow the channel count is getting set, incorrectly, to zero. Looks like the following from Dcom

```text
SEND SourceControl.ConfigureAbacoSource {"ActiveCards": [0], "AvailableCards": []}
SEND SourceControl.Start "ABACOSOURCE"
Request: {'id': 5, 'params': ['ABACOSOURCE'], 'method': 'SourceControl.Start'}

Error: PrepareRun could not run with 0 channels (expect > 0)
Could not Start Abaco
```

### LINKED ISSUE #220: Abaco source cannot keep up with data at 100 MB/sec
Possibly also there are problems in Bahama and its ability to _simulate_ data at that rate.

joefowler:
Speed up some of the compute-intensive activities involved in parsing, de-MUXing, and phase-unwrapping the Abaco data packets. Now able to run at 200+ MB/sec of Abaco data.
* Speed up parsing by replacing the calls to `binary.Read(...)`, at least for slices.
* Speed up phase unwrapping by smarter arrangement of branches and by doing each channel in a separate goroutine.
* Speed up de-multiplexing by smarter order of looping.
* Update Travis-CI to test on Ubuntu 18.04 (bionic).
Fixes #220. Fixes #218.

### DIFF
--- a/abaco.go
+++ b/abaco.go
@@ -660,6 +660,9 @@ awaitmoredata:
 				}
 				as.distributePackets(allPackets, lastSampleTime)
 			}
+			// var t1, t2 time.Time
+			// t1, t2 = t2, time.Now()
+			// fmt.Printf("Time required to distributePackets: %v\n", t2.Sub(t1))
 
 			// Align the first sample in each group. Do this by checking the first global sequenceNumber
 			// in each group and trimming leading packets if any precede the others.
@@ -682,6 +685,8 @@ awaitmoredata:
 					firstSn = sn0
 				}
 			}
+			// t1, t2 = t2, time.Now()
+			// fmt.Printf("Time required to fillMissingPackets: %v\n", t2.Sub(t1))
 
 			// For any queue that starts before maxsn0, trim the leading packets. Learn the # of samples.
 			framesToDeMUX := math.MaxInt64
@@ -695,6 +700,8 @@ awaitmoredata:
 			if framesToDeMUX <= 0 {
 				continue awaitmoredata
 			}
+			// t1, t2 = t2, time.Now()
+			// fmt.Printf("Time required to trimPacketsBefore: %v\n", t2.Sub(t1))
 
 			// Demux data into this slice of slices of RawType
 			datacopies := make([][]RawType, as.nchan)
@@ -709,6 +716,8 @@ awaitmoredata:
 				bytesProcessed += group.demuxData(dc, framesToDeMUX)
 				chanProcessed += group.nchan
 			}
+			// t1, t2 = t2, time.Now()
+			// fmt.Printf("Time required to demuxData: %v\n", t2.Sub(t1))
 
 			timeDiff := lastSampleTime.Sub(as.lastread)
 			if timeDiff > 2*as.readPeriod {
--- a/abaco.go
+++ b/abaco.go
@@ -239,9 +239,16 @@ func (group *AbacoGroup) demuxData(datacopies [][]RawType, frames int) int {
 		panic(msg)
 	}
 
-	for i, unwrap := range group.unwrap {
-		unwrap.UnwrapInPlace(&datacopies[i])
+	// Apply phase unwrapping to each channel's data. Do in parallel to use multiple processors.
+	var wg sync.WaitGroup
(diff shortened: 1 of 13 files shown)


abaco.go
```diff
@@ -201,27 +204,36 @@ func (group *AbacoGroup) demuxData(datacopies [][]RawType, frames int) int {
...
 		packetsConsumed++
+
 		switch d := p.Data.(type) {
 		case []int16:
-			// Reading vector d in order was faster than the reverse.
-			for j, val := range d {
-				idx := j % group.nchan
-				datacopies[idx] = append(datacopies[idx], RawType(val))
+			// Reading vector d in channel order was faster than in packet-data order.
+			nsamp := len(d)/nchan
```
ggggggggg, line 213: you can avoid this division by checking j against length d

packets/packets.go
```diff
@@ -401,28 +463,36 @@ func ReadPacket(data io.Reader) (p *Packet, err error) {
 			switch p.format.dtype[0] {
 			case reflect.Int16:
 				result := make([]int16, p.payloadLength/2)
-				if err = binary.Read(data, p.format.endian, result); err != nil {
+				bslice := getbytes.FromSliceInt16(result)
+				if _, err = io.ReadFull(data, bslice); err != nil {
```
ggggggggg, line 467: I think this is identical to `bytesToRawType` except special cased for `Int16`. If so, I think it is worse in the following ways
1. more verbose, it would take 2 line to do what is 5 lines here
2. I don't believe an error is possible, the code currently suggests it is
3. the current code is probably slower

packets/packets.go
```diff
@@ -484,84 +554,54 @@ type headPayloadShape struct {
...
 		case tlvTIMESTAMP: // timestamps without units
-			var x uint16
-			var y uint32
-			if err = binary.Read(data, binary.BigEndian, &x); err != nil {
-				return result, err
-			}
-			if err = binary.Read(data, binary.BigEndian, &y); err != nil {
-				return result, err
-			}
+			x := binary.BigEndian.Uint16(data[2:])
```
ggggggggg, line 583: Why are we using binary here? is it because the `getbytes` style approach only works on the native endianness, which is little? If so I think its fine. If not, I'd be in favor of switching to a getbytes approach.

abaco.go
joefowler: Yes, but that would only delay it. I still need `nsamp` at the end of the loop to advance `samplesConsumed`. And in your way, I'd need a new variable to store `len(d)` so that we don't have to call it as a function once per loop iteration. So while I don't insist that my way is better, I don't believe yours is, either.

packets/packets.go
joefowler: `bytesToRawType` does not read from an `io.Reader` like this does. So the number of lines compared is not fair, and that's also where the error is possible. Also, packets is a stand-alone sub-package that doesn't know anything about `RawType`. Finally, if `RawType` were ever compiled to be 32 bits (oh, the nightmare), then it would no longer match this case.

packets/packets.go
joefowler: It's what you said: that we need to handle big endian, which `binary` just does. That package is slow for slices, but this is actually grabbing a single `uint16` value. Honestly, I wish we weren't programming the packets to use big endian numbers in the packet headers, but we are. The firmware team basically called that a "micro optimization" and blew me off. Seems to me it would have considerably optimized MY time to use little endian, but possibly the reverse is true on the firmware end???

Sure, I could add byte swapping to `getbytes`, but that seems unlikely to be faster, and it duplicates the widely-used, correct (if slow) code that already appears in the standard `binary` package.

packets/packets.go
ggggggggg: I see, I misunderstood this, I thought you were creating the reader just to call ReadFull on it, but I think this makes sense.

ggggggggg (approved):

PR #222, opened 2020-12-04
Add group knowledge to Lancero

ggggggggg:
Fills in the info about groups with lancero. This doesn't fix the observe tab, or microscope flipping error and fb, but it does fix the number of channels and their names in microscope.


joefowler (approved):
Is this saying that the `LanceroSource` was filling the `ls.groupKeysSorted` with a placeholder value? Good idea to fix that.

PR #233, opened 2021-02-05
Unwrap reset time

### LINKED ISSUE #228: Make the phase reset length a settable parameter
It is hardcoded here: https://github.com/usnistgov/dastard/blob/e6bce83bff64b42a4b0f820744401fb97623a360/phase_unwrap.go#L29

### LINKED ISSUE #232: Dastard heartbeat ("ALIVE") message falsely says source is running after it stops
Seems like after you stop a data source, the ALIVE message incorrectly says the source is still running:

```text
STATUS    37: {'Running': False, 'SourceName': 'Abaco', 'Nchannels': 4, 'Nsamples': 4000, 'Npresamp': 1000, 'SamplePeriod': 5000, 'ChanGroups': [{'Firstchan': 0, 'Nchan': 4}], 'ChannelsWithProjectors': []}
ALIVE    38: {'Running': True, 'Time': 0.8498907080000001, 'DataMB': 1.36}
ALIVE    39: {'Running': True, 'Time': 0, 'DataMB': 0}
ALIVE    40: {'Running': True, 'Time': 0, 'DataMB': 0}
```

That doesn't seem to cause problems, but it is a bug.

joefowler:
* Allow the `PhaseUnwrapper` unwrap reset time to be set by an RPC command when the Abaco source is started.
* Fix bug where the ALIVE message said Running=true after source was stopped.
Fixes #228. Fixes #232.

### DIFF
--- a/abaco.go
+++ b/abaco.go
@@ -38,14 +38,14 @@ type AbacoGroup struct {
 }
 
 // NewAbacoGroup creates an AbacoGroup given the specified GroupIndex.
-func NewAbacoGroup(index GroupIndex) *AbacoGroup {
+func NewAbacoGroup(index GroupIndex, unwrapEnable bool, unwrapReset int) *AbacoGroup {
 	g := new(AbacoGroup)
 	g.index = index
 	g.nchan = index.Nchan
 	g.queue = make([]*packets.Packet, 0)
 	g.unwrap = make([]*PhaseUnwrapper, g.nchan)
 	for i := range g.unwrap {
-		g.unwrap[i] = NewPhaseUnwrapper(abacoFractionBits, abacoBitsToDrop)
+		g.unwrap[i] = NewPhaseUnwrapper(abacoFractionBits, abacoBitsToDrop, unwrapEnable, unwrapReset)
 	}
 	return g
 }
@@ -397,8 +397,10 @@ type AbacoSource struct {
 
 	groups map[GroupIndex]*AbacoGroup
 
-	readPeriod  time.Duration
-	buffersChan chan AbacoBuffersType
+	readPeriod      time.Duration
+	buffersChan     chan AbacoBuffersType
+	unwrapEnable    bool // whether to activate unwrapping
+	unwrapResetSamp int  // unwrap after this many samples (or never if ≤0)
 	AnySource
 }
 
@@ -409,6 +411,8 @@ func NewAbacoSource() (*AbacoSource, error) {
 	source.arings = make(map[int]*AbacoRing)
 	source.groups = make(map[GroupIndex]*AbacoGroup)
 	source.channelsPerPixel = 1
+	source.unwrapEnable = true
+	source.unwrapResetSamp = 20000
 
 	deviceCodes, err := enumerateAbacoRings()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


joefowler:
@ggggggggg ready to review?

abaco.go
```diff
@@ -397,8 +397,10 @@ type AbacoSource struct {
 
 	groups map[GroupIndex]*AbacoGroup
 
-	readPeriod  time.Duration
-	buffersChan chan AbacoBuffersType
+	readPeriod      time.Duration
+	buffersChan     chan AbacoBuffersType
+	unwrapEnable    bool // whether to activate unwrapping
```
ggggggggg, line 402: This seems redundant, the `PhaseUnwrapper` struct already stores this info, it doesn't need to also be stored in the `AbacoSource`. I guess it is done so that at line line 538 in sample you can use the values inside the `Sample` function which is part of an interface, and therefore doesn't really make sense to add these values as arguments to. So I don't see a simple solution to avoid having this property. Maybe just add a comment to that effect.

phase_unwrap.go
```diff
@@ -29,14 +30,31 @@ func NewPhaseUnwrapper(fractionBits, lowBitsToDrop uint) *PhaseUnwrapper {
 	u.lowBitsToDrop = lowBitsToDrop
 	u.twoPi = int16(1) << (fractionBits - lowBitsToDrop)
 	u.onePi = u.twoPi >> 1
-	u.resetAfter = 20000 // TODO: this should be a settable parameter
+	u.resetAfter = resetAfter
+	u.enable = enable
+	if resetAfter <= 0 {
```
ggggggggg, line 35: This is silent error handling. Do not do silent error handling. Just panic and throw an error in the relevant RPC function. Do not silently change user input parameters.

abaco.go
```diff
@@ -409,6 +411,8 @@ func NewAbacoSource() (*AbacoSource, error) {
 	source.arings = make(map[int]*AbacoRing)
 	source.groups = make(map[GroupIndex]*AbacoGroup)
 	source.channelsPerPixel = 1
+	source.unwrapEnable = true
+	source.unwrapResetSamp = 20000
```
ggggggggg, line 415: You already have these same default values in `rpc_server.go`, so I think the ones here are never used since they should immediately  be overwritten? I suggest either removing them or setting them to more obviously wrong values like `-100000`. In either case add a comment that the values are expected to be overwritten by the configure function.

ggggggggg (changes requested):

ggggggggg (changes requested):
A few minor comments, and please don't do any silent error handling. If in doubt, panic.

ggggggggg (changes requested):
A few minor comments, and please don't do any silent error handling. If in doubt, panic.

ggggggggg (changes requested):
A few minor comments, and please don't do any silent error handling. If in doubt, panic.

phase_unwrap.go
joefowler: Well, one could just as easily view this as enforcing consistency. That is, non-positive values here are taken to _mean_ disabled phase unwrapping.

However, I admit that this would not be the most obvious interpretation. I think a safer procedure might be to panic in case of inconsistency here, but to enforce consistency in the RPC server so that within Dastard the panic can't actually happen. I'll do that.

abaco.go
joefowler: Will remove.

abaco.go
joefowler: Yeah, the problem being solved here is that the `AbacoSource.Configure()` function is where we learn what the user had in mind for unwrapping enable/reset time, but the `DataSource.Sample()` function is where we need to _use_ that information. So it has to be stored in the `AbacoSource` object. I guess I can comment to this effect.

joefowler:
Okay, see what you think about the newest. If Travis CI says it's still passing our tests, then I'm ready.

ggggggggg (approved):

PR #237, opened 2021-03-11
Have Travis-CI build and deploy binaries (to GitHub)

### LINKED ISSUE #236: should we deploy binaries somehow?
I ask because we want to make an integration test that includes dastard for scan_server. I'd rather not build from source, or just include the binary in that repo. So it would be nice if dastard binaries were available somewhere we could download them from for our test.

Example of how: https://blog.questionable.services/article/build-go-binaries-travis-ci-github/

joefowler:
Deploy binaries to GitHub _automatically_. Awesome. Fixes #236.

Propose that we tag as version 0.2.9 when complete.


ggggggggg:
So cool. Thanks!

PR #243, opened 2021-05-21
Abaco UDP receiver

joefowler:
Completes UDP receiver for Abaco data source.

### DIFF
--- a/abaco.go
+++ b/abaco.go
@@ -465,8 +465,10 @@ func (device *AbacoUDPReceiver) start() (err error) {
 	go func() {
 		defer close(singlepackets)
 		buffer := make([]byte, 16384)
+		outofband := make([]byte, 16384)
 		for {
-			n, _, err := device.conn.ReadFrom(buffer)
+			// n, _, err := device.conn.ReadFrom(buffer)
+			n, _, _, _, err := device.conn.ReadMsgUDP(buffer, outofband)
 			// The source address is ignored for now.
 			if err != nil {
 				// No error printout. Error here is the normal way to detect closed connection.
--- a/abaco.go
+++ b/abaco.go
@@ -425,7 +425,6 @@ func NewAbacoUDPReceiver(hostport string) (dev *AbacoUDPReceiver, err error) {
 }
 
 // start starts the UDP device by...?
-// TODO: consider whether the newest data a *non-full* ring can be used here?
 func (device *AbacoUDPReceiver) start() (err error) {
 	raddr, err := net.ResolveUDPAddr("udp", device.host)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


abaco.go
```diff
@@ -388,17 +406,171 @@ func enumerateAbacoRings() (rings []int, err error) {
...
+	singlepackets := make(chan []byte, 20)
+	// This goroutine handles UDP message sent one at a time on singlepackets by the other goroutine.
+	// It converts them into packet.Packet objects, queues them into a slice of *packets.Packet
+	// pointers and sends the whole slice when a request comes on device.sendmore.
+	go func() {
+		defer close(device.data)
+		defer func() { device.conn = nil }()
+
+		const initialQueueCapacity = 20000
+		queue := make([]*packets.Packet, 0, initialQueueCapacity)
```
ggggggggg, line 454: This feels like a potential performance trap. We're allocating space for 20k packets repeatedly? Are we potentially only using a small fraction of that space, and thus using up much more memory than we need? What is the purpose of this sort of double buffer, why can't whatever comes next just read from single packets?

abaco.go
```diff
@@ -388,17 +406,171 @@ func enumerateAbacoRings() (rings []int, err error) {
...
+		}
+	}()
+	// This goroutine reads the UDP socket and puts the resulting packets on channel singlepackets.
+	// They will be removed and queued by the previous goroutine.
+	go func() {
+		defer close(singlepackets)
+		for {
+			message := bufferPool.Get().([]byte)
+			if _, _, err := device.conn.ReadFrom(message); err != nil {
+				// Don't report errors. Getting an error here is the normal way to detect closed connection.
```
ggggggggg, line 484: shouldn't we respond to a closed connection?Like by returning?

ggggggggg: Ohh I see we do return.... i think the comment could be more useful about what we do do instead of about what not to do.

abaco.go
```diff
@@ -435,27 +610,49 @@ func NewAbacoSource() (*AbacoSource, error) {
...
 
 // AbacoSourceConfig holds the arguments needed to call AbacoSource.Configure by RPC.
 type AbacoSourceConfig struct {
 	ActiveCards     []int
 	AvailableCards  []int
-	Unwrapping      bool // whether to activate unwrapping
-	UnwrapResetSamp int  // unwrap after this many samples (or never if ≤0)
+	HostPortUDP     []string // host:port pairs to listen for UDP packets
+	Unwrapping      bool     // whether to activate unwrapping
+	UnwrapResetSamp int      // unwrap after this many samples (or never if ≤0)
```
ggggggggg, line 622: doesnt this "reset the unwrap offset" not "unwrap" after a certain number of samples?

ggggggggg (approved):
looks good enough to merge, I noted one spot I think may be a performance trap, but otherwise nothing leapt out at me

abaco.go
joefowler: Good point. I'll change comment here and at line 576 to say "unwrap resets after this many...".

abaco.go
joefowler: Sure. The comment was about the code that's now commented out. I'll clarify.

abaco.go
joefowler: No, we're allocating space for 20k _pointers_ (to packets) repeatedly. That's going to be 160,000 bytes. We expect up to several x 10,000 packets per second. The queue will have to hold some ~50 ms of data before we pass it to the handler, which could be several thousand packets. By sizing it to a capacity of 20k at the start, we can prevent the reallocate-and-copy steps required each time it has to grow.

That "~50 ms" comes from line 818 in this function: https://github.com/usnistgov/dastard/blob/8d3358c78bc6d1f909d1bc8c2cd4c18d042dc9cc/abaco.go#L811-L821

Anyway, if we are running 250 MB/sec and 8100 payload bytes per UDP packet, then we expect only about 1500 packets per readout. Probably, then, we should change this initial capacity to 2048. If it so happens that the rate is higher, or the packets have only half the payload (as in one recent test), then maybe the queue would have to resize once, but only once.

This is a performance tuning. I believe we want this capacity to be bigger than the usual number of packets expected per readout, but the costs are small if we guess low OR high. I'll change 20000 -> 2048, but it's probably a small matter.

ggggggggg: Ok, I guess I still don't see why we do this at all, but it's at least not much memory so should be fine.

PR #244, opened 2021-05-25
define PrepareChannels for RoachSource

ggggggggg:
I basically just used the definition from abaco, but left out the channel groups. It passes tests on my macbook. Might work.


joefowler (approved):
Looks good to me. I have a few changes to the comments that I will push now. Let's merge once my changes are there (and tests are passing).

PR #252, opened 2021-10-12
Bias unwrapping

joefowler:


joefowler:
Exactly. I think the switch to uint16 was irrelevant to the operation of Dastard, but it allowed the LJH files produced by Dastard to be analyzed in the same way and by the same code as TDM-based data. Without the switch, downstream analysis code would have had to bifurcate based on the TDM/µMUX source of the data.

Issue #260, opened 2022-03-09
triggering panic while testing with triangles


ggggggggg:
I was trying to test triggering stability using triangle waves from the crate and ran into this while changing triangle parameters. At the time they were 0,6,1 dwell, steps, step size.

```panic: dsp.EdgeMultiVerifyNMonotone 66, dsp.NSamples 128, dsp.NPreSamples 64

goroutine 415274 [running]:
github.com/usnistgov/dastard.(*DataStreamProcessor).edgeMultiTriggerComputeAppend(0xc0000f4480, {0x0, 0x0, 0x0})
	/home/xf07id1/go/src/github.com/usnistgov/dastard/triggering.go:233 +0x1205
github.com/usnistgov/dastard.(*DataStreamProcessor).TriggerData(0xc0000f4480)
	/home/xf07id1/go/src/github.com/usnistgov/dastard/triggering.go:573 +0x4a
github.com/usnistgov/dastard.(*DataStreamProcessor).processSegment(0xc0000f4480, 0xc0004782a0)
	/home/xf07id1/go/src/github.com/usnistgov/dastard/process_data.go:118 +0x46
github.com/usnistgov/dastard.(*AnySource).ProcessSegments.func1(0x47059a8000000080)
	/home/xf07id1/go/src/github.com/usnistgov/dastard/data_source.go:347 +0x58
created by github.com/usnistgov/dastard.(*AnySource).ProcessSegments
	/home/xf07id1/go/src/github.com/usnistgov/dastard/data_source.go:345 +0x93
(qsp) [xf07id1@xf07id-det3 ~]$ 
```

joefowler:
So this is not the `TriangleSource` but a Lancero source where the data are triangles? 

This must be related to #259 #257 #256 and #254 and to our plans for fixing them, because the basic edge multi trigger (EMT) design seems to have unhappy edge cases. We have discussed removing the multi-trigger aspect entirely (still considering).

Issue #259, opened 2022-03-08
needed edgeMultiInternalSearchState reset to initial prints A LOT


ggggggggg:
https://github.com/usnistgov/dastard/blob/a573d0d1248efe9a55ce490fec33e624af76a59d/triggering.go#L207

Often right after I send triggers this will print like 10x per channel then stop.

Issue #256, opened 2022-03-02
reset edgeMulti triggering on data drop


ggggggggg:
On data drop, we have discontinuous data, and edgeMulti doesn't like that. So we need to reset it on data drop, this should fix the panics Jamie is seeing.

PR #264, opened 2022-03-14
Trigger refactor

### LINKED ISSUE #255: do triggering with structured concurrency
Refactor how triggering is done to eliminate the sue of go channels to the extent possible. Instead split `processSegments` into two functions, add another waitgroup loop that does triggering. Then do the group triggering in a function after that waitgroup loop.

joefowler:
Complete redesign of the triggering logic and especially the `TriggerBroker`, from which we are removing the long-running goroutine and all channels for its input and output. The code is no shorter than before, but we hope it should be clearer. It uses concurrency only where performance requires it (spinning a goroutine for each channel's data processing) and uses the classic simple concurrency model where all _N_ concurrent processes are waited on before proceeding. Fixes #255.

### DIFF
--- a/edge_multi_test.go
+++ b/edge_multi_test.go
@@ -117,7 +117,7 @@ func TestEdgeMultiParts1(t *testing.T) {
 				{firstRisingFrameIndex: 12, npre: 2, nsamp: 4},
 				{firstRisingFrameIndex: 14, npre: 2, nsamp: 4},
 				{firstRisingFrameIndex: 16, npre: 2, nsamp: 4}},
-			"edgeMultiComputeAppendRecordSpecs: 2 records far enough apart to trigger both, negative going",
+			"edgeMultiComputeAppendRecordSpecs: lots over overlapping full length records",
 		},
 	}
 	for _, test := range tests {


group_trigger.go
```diff
@@ -200,20 +190,21 @@ func (broker *TriggerBroker) Distribute(primaries map[int]triggerList) (map[int]
 	for idx, tlist := range primaries {
 		broker.latestPrimaries[idx] = tlist.frames
 		nprimaries += len(tlist.frames)
-		err := broker.triggerCounters[idx].observeTriggerList(&tlist)
+		err := broker.triggerCounters[idx].countNewTriggers(&tlist)
 		if err != nil {
 			log.Printf("triggering assumptions broken!\n%v\n%v\n%v", err,
```
ggggggggg, line 195: I was going to suggest changing this panic to just log, which it seems you did.... so good

joefowler: Don't think this was ever a panic, at least not recently.

ggggggggg:
Is it getting send once per channel per second instead of once per second? I think it was sent at the right rate before.

ggggggggg:
fixes #260 
fixes #259 
fixes #256 
fixes #255

edge_multi_test.go
```diff
@@ -109,46 +60,97 @@ func TestEdgeMultiParts1(t *testing.T) {
...
+				nmonotone: 1, npre: 2, nsamp: 4},
+			[]RecordSpec{{firstRisingFrameIndex: 2, npre: 2, nsamp: 4},
+				{firstRisingFrameIndex: 4, npre: 2, nsamp: 4},
+				{firstRisingFrameIndex: 6, npre: 2, nsamp: 4},
+				{firstRisingFrameIndex: 8, npre: 2, nsamp: 4},
+				{firstRisingFrameIndex: 10, npre: 2, nsamp: 4},
+				{firstRisingFrameIndex: 12, npre: 2, nsamp: 4},
+				{firstRisingFrameIndex: 14, npre: 2, nsamp: 4},
+				{firstRisingFrameIndex: 16, npre: 2, nsamp: 4}},
+			"edgeMultiComputeAppendRecordSpecs: 2 records far enough apart to trigger both, negative going",
```
ggggggggg, line 120: this error string is not accurate

PR #265, opened 2022-03-16
merge both trigger changes

ggggggggg:
refactor edge multi triggering to be
1. easier to read/reason about
2. less state carried over between data segments, in particular no index is carried over so we're less likely to index out of bounds
3. panic free
I also made a few changes suggested by the linter, but not all, to reduce the size of the diff.


triggering_edge_multi_low_level_test.go
```diff
@@ -33,23 +33,23 @@ func TestEdgeMultiParts1(t *testing.T) {
 	assert.Equal(t, FrameIndex(12), s.v, "EMTState should have u==v, to indicated that u has has been recordized")
 
 	// two records far enough apart to trigger both
-	raw_A := [21]RawType{0, 0, 0, 0, 10, 20, 0, 0, 10, 20, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
```
ggggggggg: The underscore was useful to make it easy to copy and paste then select all `_A` and change all to `_B`. This change is ok though.

joefowler (approved):

PR #268, opened 2022-03-17
Group trigger is ready!

### LINKED ISSUE #247: Group triggering cannot be enabled
This is a feature we never completed. Although there's a `TriggerBroker` in [group_trigger.go](https://github.com/usnistgov/dastard/blob/master/group_trigger.go), we don't have yet any RPC commands to activate its features. HEATES will need it.
* Add RPC commands to manipulate the trigger broker.
* Store configuration of group trigger.
* Make it work.
* Fix #235 (I assume this is related?)
* In Dastard-commander, make GUI elements that can do elementary group triggering.

joefowler:
Fixes #247 finally. Features:
- Group triggers can connect any channel to any other.
- Request to trigger from X to itself is ignored.
- RPC commands to add and remove specific connections, or to remove all connections.
- The new trigger architecture (#264) is used.
- BAHAMA data simulator has a "crosstalk" mode where odd-numbered channels are 2% the usual level (useful for testing group trigger).


ggggggggg (approved):
lgtm

PR #269, opened 2022-03-18
Use new (go 1.17+) approach to change slices' types

joefowler:
For go 1.17+ the `unsafe.Slice(...)` [function](https://pkg.go.dev/unsafe#Slice) is the recommended way to convert the type of elements of a slice without actually changing the underlying data. See [this playground example](https://go.dev/play/p/GMWsdLHK6Zv?v=goprev), which shows that the data of the original and converted slices live at the same location, and that changing the data in one changes it in the other.

Use conditional compilation to use this for Go 1.17+, and to use the old (now deprecated) versions for Go -1.16. Sweet!


ggggggggg (approved):
we should do the same in getbytes. i think the reason at least some of these aren't in getbytes is that they work on RawType.... seems like it might be an issue if we ever change the type of RawType

joefowler:
Good point about the getbytes package. Hmm. Let's delay this PR until after spring break. Then I will
- [ ] Also switch `getbytes` to use conditional compilation and (when Go 1.17+) `unsafe.Slice`
- [ ] Remove `bytesToInt32()`, because I think it's unused (other than tests). If I'm wrong, put it in the `getbytes` package instead.

Issue #271, opened 2022-05-13
Dastard crashes with high memory usage if config is present


cjtitus:
I'm on emtfix2 branch, commit a0aaa86e10e9f8

When I start Dastard, then Dcom, and start data, if I have a config file present, Dastard basically runs my computer out of memory, grinds to a halt, and then dcom crashes due to missed heartbeats. Dastard itself doesn't always crash, but I need to kill it to keep it from taking out my PC.

I have attached a config file that caused a crash at startup. However, as this config file was necessarily the result of a previous, working configuration, I have no reason to actually suspect that the specific configuration is the problem. Rather, starting Dastard with any config seems to be the problem.
[config_pathological.txt](https://github.com/usnistgov/dastard/files/8688946/config_pathological.txt)

ggggggggg:
I'll look at this, should be fairly straightforward.

joefowler:
Fixed by #272.

PR #272, opened 2022-05-13
Fixconfigcrash

ggggggggg:
fixes a crash when EdgeMulti is saved as true but the EMT state is not saved

also fixes some EMT crashes

### DIFF
--- a/abaco_test.go
+++ b/abaco_test.go
@@ -78,7 +78,7 @@ func TestGeneratePackets(t *testing.T) {
 	}
 }
 
-func _skip_TestAbacoRing(t *testing.T) {
+func TestAbacoRing(t *testing.T) {
 	if _, err := NewAbacoRing(99999); err == nil {
 		t.Errorf("NewAbacoRing(99999) succeeded, want failure")
 	}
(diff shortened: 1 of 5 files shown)


abaco_test.go
```diff
@@ -78,7 +78,7 @@ func TestGeneratePackets(t *testing.T) {
 	}
 }
 
-func _skip_TestAbacoRing(t *testing.T) {
+func TestAbacoRing(t *testing.T) {
```
joefowler, line 81: I don't understand why these were skipped at all. Did you commit to Master without me noticing? In my version of master they aren't skipped.

joefowler (approved):
If this fixes #271, then great. I'm pushing a change to update the Dastard version number to 0.2.13 and add a comment to `VERSION.md`.

abaco_test.go
ggggggggg: I must have. They crash a lot, so when I was running tests in the debugger, I had to skip them to actually catch the crashes I was interested in in the EMT code.

joefowler: Bad Galen! Good for restoring them.

ggggggggg: No, bad tests!!! Brilliant skipping to make progress!

PR #274, opened 2022-05-26
Unwrap dropbits option

ggggggggg:


ggggggggg:
The tests failed, but I don't think it's real. It's just a segfault in the `Abaco` tests, which is why I always skip the abaco tests! But this change is about abaco, so doesn't seem good to skip.  
image

ggggggggg:
So there is a real failure in the `packet_test.go`, probably because I changed how timestamps without units from tlvs are interpreted. I can't understand the intention of how they were supposed to behave, so my inclination is to just change the tests to pass.

```
--- FAIL: TestTLVs (0.00s)
    packets_test.go:241: PacketTimestamp = {217304205466533888 1e+09}, want {3315799033608 0}
```

Do you want to look at it before I do that and see if you have a different opinion on how I should deal with it?

packets/packets.go
```diff
@@ -579,7 +580,15 @@ func parseTLV(data []byte) (result []interface{}, err error) {
 		case tlvTIMESTAMP: // timestamps without units
 			x := binary.BigEndian.Uint16(data[2:])
 			y := binary.BigEndian.Uint32(data[4:])
-			result = append(result, MakeTimestamp(x, y, 0.0))
+
+			// we assume a 64 bit int with units of nanoseconds
```
joefowler, line 584: I guess I can approve this, but I'm a little concerned that we'll be making life more difficult in the future. The hope is that no timestamp source that has any meaningful units would use the `tlvTIMESTAMP` type but would always favor `tlvTIMESTAMPUNIT`. 

What was wrong with the previous choice to set `PacketTimestamp.Rate=0`? Did something break when we did that?

joefowler: (I could totally imagine that Rate=0 breaks something.)

abaco.go
```diff
@@ -45,25 +45,45 @@ type AbacoGroup struct {
...
+	PulseSign  int  // direction data will go when pulse arrives, used to calculate bias level
+}
+
+func (u AbacoUnwrapOptions) isvalid() error {
+	if u.Enable && !u.DropBits {
+		return fmt.Errorf("should not have both enable and DropBits true in AbacoUnwrapOpts")
+	}
+	return nil
+}
+
```
joefowler, line 62: If you remove the `DropBits` field like I propose (see other comments), then the parameters are more automatically self-consistent, and this entire test becomes unnecessary.

abaco.go
```diff
@@ -626,17 +642,17 @@ func (as *AbacoSource) Delete() {
...
-	AvailableCards  []int
-	HostPortUDP     []string // host:port pairs to listen for UDP packets
-	Unwrapping      bool     // whether to activate unwrapping
-	UnwrapResetSamp int      // unwrap resets after this many samples (or never if ≤0)
-	PulseSign       int      // should be one of (-1,+1) to choose the reset point
-	Bias            bool     // whether to use biased unwrapping
+	ActiveCards    []int
+	AvailableCards []int
+	HostPortUDP    []string // host:port pairs to listen for UDP packets
+	AbacoUnwrapOptions
```
joefowler, line 648: Ooh! Nice use of "composition" in Golang.

joefowler (changes requested):
This looks mostly good to me. You will see that I believe that the `DropBits` boolean is extraneous and should be removed. Let me explore that proposal and see if it works. I'll start from the PR and make sure that I can remove that field.

joefowler (approved):

PR #275, opened 2022-06-01
Unwrap dropbits option take 2

ggggggggg:
Trying it Joe's way. It was definitley more complicated overall, but it works and thats good enough for me. We should merge this and close the other PR.


joefowler:
We did this a slightly different way (#274), so I'm closing this stale PR.

PR #278, opened 2022-09-15
handle trimming in only one spot, also always use stream.x instead of…

ggggggggg:
1. call `DataStream.TrimStream` in only one location, with correct value
2. prefer `stream.x` over `stream.segment.x` or `segment=stream.segment;segment.x` for clairity

### DIFF
--- a/edge_multi_trigger.go
+++ b/edge_multi_trigger.go
@@ -360,7 +360,7 @@ func (dsp *DataStreamProcessor) edgeMultiTriggerComputeAppend(records []*DataRec
 	// fmt.Println("firstFrameIndex", stream.firstFrameIndex, "len(segment.rawData)", len(stream.rawData))
 	recordSpecs := dsp.EMTState.edgeMultiComputeRecordSpecs(stream.rawData, stream.firstFrameIndex)
 	for _, recordSpec := range recordSpecs {
-		record := dsp.triggerAtSpecificSamples(&stream, int(recordSpec.firstRisingFrameIndex-stream.firstFrameIndex), int(recordSpec.npre), int(recordSpec.nsamp))
+		record := dsp.triggerAtSpecificSamples(int(recordSpec.firstRisingFrameIndex-stream.firstFrameIndex), int(recordSpec.npre), int(recordSpec.nsamp))
 		records = append(records, record)
 	}
 	// n_before := len(stream.rawData)
--- a/edge_multi_trigger.go
+++ b/edge_multi_trigger.go
@@ -285,8 +285,9 @@ func edgeMultiShouldRecord(t, u, v FrameIndex, npreIn, nsampIn int32, mode EMTMo
 }
 
 // NToKeepOnTrim returns how many samples to keep when trimming a stream.
-// GCO thinks the minimum is s.nsamp + 4 based on the lookback in the kink model, but
-// I made the function keep more just to be conservative and it doesn't
+// GCO thinks the minimum is (s.nsamp + 4) based on the lookback in the kink model, but
+// the function keeps more just to be conservative (so we can experiment with
+// slight changes in trigger algorithm in the future).
 func (s EMTState) NToKeepOnTrim() int {
 	return int(2*s.nsamp + 10)
 }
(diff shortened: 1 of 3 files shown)


edge_multi_trigger.go
```diff
@@ -316,9 +318,11 @@ func (s *EMTState) edgeMultiComputeRecordSpecs(raw []RawType, frameIndexOfraw0 F
 			break
 		}
 		t, u, v = u, v, FrameIndex(x.triggerInd)+frameIndexOfraw0
-		// fmt.Println(t, u, v)
 		recordSpec, valid := edgeMultiShouldRecord(t, u, v, s.npre, s.nsamp, s.mode)
 		if valid {
+			// if (recordSpec.firstRisingFrameIndex - frameIndexOfraw0) < FrameIndex(s.npre) {
+			// 	panic("wtf")
+			// }
```
joefowler, line 325: There are several places like this one, where you added but commented out some code. Probably should remove it all.

edge_multi_trigger.go
```diff
@@ -355,12 +362,16 @@ func (s *EMTState) edgeMultiComputeRecordSpecs(raw []RawType, frameIndexOfraw0 F
...
+	recordSpecs := dsp.EMTState.edgeMultiComputeRecordSpecs(stream.rawData, stream.firstFrameIndex)
 	for _, recordSpec := range recordSpecs {
-		record := dsp.triggerAtSpecificSamples(segment, int(recordSpec.firstRisingFrameIndex-segment.firstFrameIndex), int(recordSpec.npre), int(recordSpec.nsamp))
+		record := dsp.triggerAtSpecificSamples(&stream, int(recordSpec.firstRisingFrameIndex-stream.firstFrameIndex), int(recordSpec.npre), int(recordSpec.nsamp))
 		records = append(records, record)
 	}
-	dsp.stream.TrimKeepingN(dsp.EMTState.NToKeepOnTrim()) // see comment at end of edgeMultiComputeAppendRecordSpecs
+	// n_before := len(stream.rawData)
+	// n_keep := dsp.EMTState.NToKeepOnTrim()
+	// dsp.stream.TrimKeepingN(n_keep) // see comment at end of edgeMultiComputeAppendRecordSpecs
```
joefowler, line 374: This is a `TrimKeepingN` instance that you have removed. Can you explain the thinking? Was edge-multi trimming streams prematurely? How did this ever work...? Did this make EMT play badly with all other triggers (as noted on the Dcom GUI)?

joefowler: Actually, I think my problem reviewing this PR is that I don't know what problem you're trying to fix...??

triggering.go
```diff
@@ -35,22 +35,22 @@ type TriggerState struct {
 }
 
 // create a record using dsp.NPresamples and dsp.NSamples
-func (dsp *DataStreamProcessor) triggerAt(segment *DataSegment, i int) *DataRecord {
-	record := dsp.triggerAtSpecificSamples(segment, i, dsp.NPresamples, dsp.NSamples)
+func (dsp *DataStreamProcessor) triggerAt(stream *DataStream, i int) *DataRecord {
```
joefowler, line 38: I guess we can do this. But what's the pedantic argument?

joefowler (changes requested):

triggering.go
ggggggggg: it's to be conceptually consistent. The DSP is a `DataStreamProcessor`. It makes sense that it processes a `DataStream`. It's confusing when it sometimes handles a `DataStream` and other time handles a `DataSegment`. One spends mental effort trying to figure out why one function takes one and another takes the other. This way they all just take `DataStream` and you don't need to wonder why you would choose `Stream` or `Segment` in any part of the code.

edge_multi_trigger.go
ggggggggg: The issue is dastard would crash every time i started edge multi triggers. I have no idea how it ever worked.

joefowler: Okay, so this fixes #279 (I just added that as an issue).

triggering.go
joefowler: This makes sense.

joefowler (approved):
I simplified the `triggerAt` and `triggerAtSpecificSamples` so they no longer need their first argument. We debated whether that should be called a stream or a segment, but that argument is *always* `dsp.stream`, so there's no point in passing it. It seems clearer to me without.

ggggggggg:
fixes #279

PR #280, opened 2022-09-16
Fix problem with crashing demo data when rate is very low.

### LINKED ISSUE #277: triggerBroker/triggerCounter crash
A crash can occur when changing the number of channels that are triggering that has to do with counting the number of triggerCounter messages. The linked code below, line 224, is what causes the number of messages to vary.

https://github.com/usnistgov/dastard/blob/bda5d615efba30fc7a3a8372e957c49abbbfa7d4/group_trigger.go#L220-L229

Probably instead of using maps here, we should use arrays and always iterate over all channels? Then there would be not counting step, because it would always just be the length of the array.

joefowler:
Fixes #277, a problem found when the rate of (typically simulated) data is unusually low.

### DIFF
--- a/group_trigger.go
+++ b/group_trigger.go
@@ -76,6 +76,9 @@ func (tc *TriggerCounter) messageAndReset() {
 	tc.hi = tc.keyFrame + FrameIndex(roundint(tc.sampleRate*hi_minus_key))
 }
 
+// countNewTriggers increments the relevant per-channel counters.
+// It also generates a set of messages in `tc.messages` at the
+// chosen message rate (i.e., each `tc.stepDuration`).
 func (tc *TriggerCounter) countNewTriggers(tList *triggerList) error {
 	// Update keyFrame and keyTime to have a new, recent correspondence between
 	// the real-world time and frame number.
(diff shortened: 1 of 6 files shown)


group_trigger.go
```diff
@@ -81,22 +82,22 @@ func (tc *TriggerCounter) countNewTriggers(tList *triggerList) error {
 	tc.keyFrame = tList.keyFrame
 	tc.keyTime = tList.keyTime
 	tc.sampleRate = tList.sampleRate
+	if tc.sampleRate <= 0 {
```
ggggggggg, line 85: Why return nil here? shouldnt negative sample rate be indicative of an error or some sort?

group_trigger.go
```diff
@@ -81,22 +82,22 @@ func (tc *TriggerCounter) countNewTriggers(tList *triggerList) error {
...
+		// Counting trigger rates
+		return nil
+	}
 	if !tc.initialized {
 		tc.initialize()
 	}
 	for _, frame := range tList.frames {
-		if frame > tc.hi {
+		for frame > tc.hi {
```
ggggggggg, line 93: a comment noting that `tc.hi` will be incremented by `messageAndReset` makes this look less like an infinite loop

group_trigger.go
```diff
@@ -81,22 +82,22 @@ func (tc *TriggerCounter) countNewTriggers(tList *triggerList) error {
...
-		}
-		if frame < tc.lo {
-			return fmt.Errorf("countNewTriggers: observed count before lo=%v, frame=%v", tc.lo, frame)
+		if frame >= tc.lo {
+			tc.countsSeen++
 		}
-		tc.countsSeen++
 	}
-	if tList.firstFrameThatCannotTrigger > tc.hi {
+	for tList.firstFrameThatCannotTrigger > tc.hi {
```
ggggggggg, line 100: a similar commend about `messageAndReset` increasing `tc.hi` would be helpful

ggggggggg (approved):
a few minor requests for comments in the code, but basically seems fine

group_trigger.go
joefowler: Good point. I'll do that.

group_trigger.go
joefowler: This fixed some strange error I was having that arose from messages being generated in a triangle or other fake-data sample, where there was no sample rate. I just decided that having a zero sample rate was not a problem that needed to be diagnosed (or panicked at) here in the group trigger.

PR #282, opened 2022-09-21
Fix go lint/warnings and deprecated libraries

joefowler:


joefowler (commented):
`io/ioutil` is deprecated, and all the functionality is elsewhere (in `os` or in `io`, since go 1.16). Also fixed some warnings about redundancy and about how you should check errors before a `defer` statement to release a resource.

PR #284, opened 2022-09-23
Change zmq library to pebbe/zmq4, which also wraps cgo+czmq 

joefowler:
This still needs more live tests, but a few quick ones on a working µMUX system makes it look promising. Worth reviewing now.


ggggggggg (commented):
looks good for testing, but lets test before merge

Issue #290, opened 2022-09-30
Add a main program to read the first N UDP packets and dump their headers


joefowler:
Diagnosing problems with packets can be tricky. Let's add a small main program that reads the first N UDP packets and dumps useful info about their header (maybe also contents?). Later, could have it read a shared-memory ring buffer, if that is useful.

PR #291, opened 2022-09-30
Packet dumper, a new program

joefowler:
This adds `cmd/udpdump/udpdump.go`, a separate main program that watches for Abaco-style packets from the UDP source specified at the command line. It prints key header information for the first N packets received.


joefowler:
Fixes #290 when merged

ggggggggg (approved):

PR #294, opened 2022-11-18
Faster dastard

joefowler:
The work of speeding up Dastard for µMUX systems is still in progress, but I am pretty sure we want this set of changes incorporated. It does _not_ include the work to trigger-then-unwrap, which is speculative and can be found in branch:unwrap_records.

Changes:
* Benchmark test for Bahama production of UDP packets.
* Change design of main Abaco loop to reduce number of parallel steps.
* Allow command-line arguments to run CPU or memory profiling.
* Better documentation in code and README.
* Bump version number.


joefowler (commented):
I think it's good to get these changes onto master.

PR #296, opened 2022-11-22
Overlap record bug

joefowler:
Fix bug in legacy trigger that causes overlapping records. Is tested on 687medusa.

### DIFF
--- a/triggering.go
+++ b/triggering.go
@@ -113,7 +113,7 @@ func (dsp *DataStreamProcessor) levelTriggerComputeAppend(records []*DataRecord)
 	for i := dsp.NPresamples; i < ndata+dsp.NPresamples-dsp.NSamples; i++ {
 
 		// Now skip over 2 record's worth of samples (minus 1) if an edge trigger is too soon in future.
-		// Notice how this works: edge triggers get priority, vetoing (1 record minus 1 sample) into the past
+		// Existing edge triggers get priority, vetoing (1 record minus 1 sample) into the past
 		// and 1 record into the future.
 		if FrameIndex(i)+nsamp > nextFoundTrig {
 			i = int(nextFoundTrig) + dsp.NSamples - 1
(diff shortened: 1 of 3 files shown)


triggering.go
```diff
@@ -188,49 +206,41 @@ func (dsp *DataStreamProcessor) autoTriggerComputeAppend(records []*DataRecord)
...
-		trigList.keyFrame = dsp.stream.DataSegment.firstFrameIndex
-		trigList.keyTime = dsp.stream.DataSegment.firstTime
-		trigList.sampleRate = dsp.SampleRate
-		trigList.firstFrameThatCannotTrigger = dsp.stream.DataSegment.firstFrameIndex +
-			FrameIndex(len(dsp.stream.rawData)) - FrameIndex(dsp.NSamples-dsp.NPresamples)
-
-		dsp.lastTrigList = trigList
-		// EdgeMulti does not play nice with other triggers, so return now!!
-		return records
-	}
```
joefowler, line 211: To be clear, the block above is completely redundant from lines 196-209. That was done to make it safe to return, but I didn't like the duplication. I simplified the logic, in my opinion.

triggering_test.go
```diff
@@ -587,20 +636,40 @@ func BenchmarkLevelTrigger0TriggersOpsAreSamples(b *testing.B) {
...
 
+func almostEqual(a, b, threshold float64) bool {
+	return math.Abs(a-b) <= threshold
+}
+
 func TestKinkModel(t *testing.T) {
 	xdata := []float64{0, 1, 2, 3, 4, 5, 6, 7}
 	ydata := []float64{0, 0, 0, 0, 1, 2, 3, 4}
-	ymodel, a, b, c, X2, err := kinkModelResult(3, xdata, ydata)
-	if a != 0 || b != 0 || c != 1 || X2 != 0 || err != nil {
```
joefowler, line 594: This test was failing ONLY on my new M1 Mac, because it was testing for exact floating-point equality. Always a bad idea, but only now did it cause any problems. Anyway, I replaced that with tests for approximate equality `almostEqual(...)`, and now it passes on all computers.

triggering_test.go
```diff
@@ -587,20 +636,40 @@ func BenchmarkLevelTrigger0TriggersOpsAreSamples(b *testing.B) {
...
+		}
+		if fail || err != nil {
+			t.Errorf("offset %.0f: a %v, b %v, c %v, X2 %v, err %v, ymodel %v",
+				test.offset, a, b, c, X2, err, ymodel)
+		}
 	}
+
 	kbest, X2min, err := kinkModelFit(xdata, ydata, []float64{1, 2, 2.5, 3, 3.5, 4, 5})
-	if kbest != 3 || X2min != 0 || err != nil {
+	if kbest != 3 || math.Abs(X2min) > 1e-15 || err != nil {
```
joefowler, line 672: Should use `almostEqual` here.

triggering.go
```diff
@@ -55,6 +55,25 @@ func (dsp *DataStreamProcessor) triggerAtSpecificSamples(i int, NPresamples int,
...
+			mindelay = autoDelaySamples
+		}
+	}
+	nextPotentialTrig := int(dsp.LastTrigger-dsp.stream.firstFrameIndex) + mindelay
+	if nextPotentialTrig < dsp.NPresamples {
+		return dsp.NPresamples
+	}
+	return nextPotentialTrig
+}
+
```
joefowler, line 76: The new function `firstPotentialTriggerFrame` is here so that we can keep track of where triggering might be able to start in a new segment. The answer, of course, depends on whether we want to ask when _auto_ triggers might happen, or other types, because auto-triggers (potentially) have an extra delay beyond the usual no-overlap requirement.

triggering_test.go
joefowler: Done in 258c54f

triggering.go
```diff
@@ -147,21 +168,18 @@ func (dsp *DataStreamProcessor) autoTriggerComputeAppend(records []*DataRecord)
...
 	nextFoundTrig := FrameIndex(math.MaxInt64)
 	if nFoundTrigs > 0 {
 		nextFoundTrig = records[idxNextTrig].trigFrame - stream.firstFrameIndex
 	}
 
-	// dsp.LastTrigger stores the frame of the last trigger found by the most recent invocation of TriggerData
-	nextPotentialTrig := dsp.LastTrigger - stream.firstFrameIndex + delaySamples
-	if nextPotentialTrig < npre {
-		nextPotentialTrig = npre
+	const auto = true
```
ggggggggg, line 178: use same name `isAuto` as other locations

Issue #300, opened 2023-01-31
Add a pulse-veto to auto triggers, or something like MATTER's "noise triggers"


joefowler:
HOLMES will have embedded radioisotope sources inside the sensors. There will be no way to "remove the source" for noise characterization. They need a way to collect noise data that vetos the record when a pulse is observed.

The simplest approach would be to add a parameter to the existing auto triggers (i.e., the triggers that repeat indefinitely on a fixed time delay of at least 1 record length). The parameter might be nothing more than a requirement that the (max-min) in a record not exceed this level.

See also the GUI issue https://github.com/usnistgov/dastardcommander/issues/132 to control the new feature.

joefowler:
Well, when one runs auto triggers, you either turn off the level+edge triggers, or you get those triggers mixed into your output stream. We don't want the latter, so the level or edge have to be off.

I believe that in MATTER (or its predecessor, the xcaldaq_client), you could turn ON edge triggers and then not allow them to be written to disk. That's not in Dastard now, but it would solve the problem the way you proposed.

Either way, though, we're adding a new feature to DASTARD: either
* you can have a type of trigger be ON but not written to disk, or
* we add a way to veto auto-triggers.

In my opinion, the latter is simpler, clearer, and a smaller change to the current design.

joefowler:
Fixed by #301.

PR #301, opened 2023-02-02
Auto trigger veto

joefowler:
Add small feature to allow auto triggers to be vetoed whenever the (max-min) is too large.


joefowler:
Fixes #300

ggggggggg (approved):
didn't look that closely, but no red flags were found

PR #312, opened 2023-03-08
Build ZMQ in github actions

ggggggggg:


ggggggggg:
github actions times out `make` since it takes longer than 3 minutes

PR #319, opened 2023-04-20
Panic when lancero devices fail at key setup points

### LINKED ISSUE #316: Make lancero device setup errors produce panics
We are calling `adapter.allocateRingBuffer` without checking its return value for errors. This causes some serious problems. In fact, the changes we made in #258 and #263 turn out to be disallowed by the design of the lancero _driver_, and failure to check this error means we had a lot of difficulty diagnosing this. Steps include:

- [x] Undo #263--change the lancero buffer back to 32 MB (had been 256 MB, but 40 MB was the allowed max).
- [x] Check return values somewhere in the stack, and panic somewhere if there was an error setting up the devices.
- [x] Look for similar failures that should cause absolute panics in the lancero device system.

joefowler:
Also change the lancero.adapter so it has high verbosity at first (in the sample-card phase) and zero verbosity during the regular data-acquisition phase.

Fixes #316.

### DIFF
--- a/lancero/lancero.go
+++ b/lancero/lancero.go
@@ -91,7 +91,11 @@ func NewLancero(devnum int) (*Lancero, error) {
 	lan.device = dev
 	lan.collector = &collector{device: dev, simulated: false}
 	lan.adapter = &adapter{device: dev, verbosity: 3}
-	lan.ChangeRingBuffer(256000000, 128000000)
+
+	OneMB := 1024 * 1024
+	if err := lan.ChangeRingBuffer(32*OneMB, 16*OneMB); err != nil {
+		panic(err)
+	}
 
 	lan.adapter.status()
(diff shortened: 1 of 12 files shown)


lancero/lancero.go
```diff
@@ -112,6 +149,11 @@ func (lan *Lancero) CollectorConfigure(linePeriod, dataDelay int, channelMask ui
 
 // StartCollector starts the data serializer.
 func (lan *Lancero) StartCollector(simulate bool) error {
+	// Panic if program was compiled with Go 1.20+, b/c that's incompatible with the Lancero device
```
ggggggggg, line 152: Shouldn't we only need this in one location? Of Start Adapter and Start Collector, isn't one strictly required for the other?

ggggggggg (approved):

PR #323, opened 2023-05-11
Rawdata archive server

joefowler:
Add an RCP call `StoreRawDataBlock(N)` to request that (at least) N samples per channel be stored to a temporary file.
* The call returns the name of the file where the data will appear when the request is complete.
* The file is an npz (numpy compressed archive) file.
* File will contain _at least_ N samples from each channel. It is not an error for there to be more than N samples.
* The raw data are in arrays named "chan34" and "err32" and similar.
* An array "externalTriggerRowcount" (possibly empty) tells when the external trigger was recorded.
* An array "firstFrameIndex" gives the frame index number of the first sample for each channel. (I think it's an error if any values are different from any others!)

This service is intended to replace the use of continuous auto-triggers and transporting them over a ZMQ PUB channel to Python. That approach lost packets (records) AND required changing trigger configurations.


joefowler:
This morning, I merged in the changes from the workaround to the Go 1.20/Lancero problem. I think this PR is 100% up to date now.

PR #339, opened 2024-02-09
Add subframe timing info to LJH and OFF file headers

### LINKED ISSUE #337: LJH 2 files should explicitly indicate the subframe divisions per frame
In the past, we've assumed that the number of TDM rows is the number of subframe divisions in the places where they are relevant:
1. In the LJH per-pulse timing info, there's a "row count" value. It's equal to frame counts times the number of rows, plus this channel's row number. That makes sense for TDM data! But for other data, not so much.
2. In the external trigger information, there's a similar assumption that our counts correspond to a frame divided into Nrows subdivisions. Again, that's only strictly required for TDM data. For µMUX data, the subdivision is nearly arbitrary! I chose 64 last week.

Proposal:
* Explicitly track the number of subframe divisions. This might not equal the number of rows in a µMUX data source (because "rows" there is an artificial construct).
* Even better, we might be able to restrict the language of "rows" only to TDM data sources (Lancero).
* Explicitly track the number of subframe steps by which this channel is offset from the exact frame start. For Lancero sources, this will be the row number; for others, it will be zero.
* Put new info into the LJH2 and OFF file headers to indicate the "Number of subframe divisions" and "Subframe delay for this channel". If not found, MASS can assume these values default to the number of rows and the row number.

joefowler:
Add two new fields to the LJH and OFF file headers: "Subframe offset" and "Subframe divisions" to describe the timing of this channel w.r.t. the overall frame clock. Expect offset=row number for Lancero source and =0 for all other sources; subframe divisions = number of rows for Lancero and arbitrary for all other sources (but must match how external triggers are timed).
Throughout code base, update the vocabulary (comments, variable names, function names) to reflect the new usage.
Fixes #337.

### DIFF
--- a/abaco.go
+++ b/abaco.go
@@ -19,9 +19,9 @@ import (
 )
 
 const abacoSubsampleDivisions = 64 // The external triggers will be resolved this much finer than the frame rate
-// That is, we are writing external trigger info at a rate that we'd call the "row rate" in a TDM system.
-// So we write the data as if Abaco data sources had a 64-row readout, but all we mean by that is that
-// they are being resolved at 64x finer time steps than the frame rate.
+// That is, we are writing external trigger info at a clock rate that we'd call the "row rate" in a TDM system.
+// So we write the ext trigger data as if Abaco data sources had a 64-row readout, but all we mean by that is
+// that they are being resolved at 64x finer time steps than the frame rate.
 
 const abacoFractionBits = 16 // changed from 13 to 16 in Jan 2021.
 const abacoBitsToDrop = 4
@@ -38,9 +38,9 @@ func gIndex(p *packets.Packet) GroupIndex {
 
 // FrameTimingCorrespondence tracks how we convert timestamps to FrameIndex
 type FrameTimingCorrepondence struct {
-	CountsPerRow  uint64
-	LastTimestamp packets.PacketTimestamp
-	LastRowCount  FrameIndex
+	SubframeRate      uint64 // Ratio of timestamp rate to subframe rate
+	LastTimestamp     packets.PacketTimestamp
+	LastSubframeCount FrameIndex
 }
 
 //------------------------------------------------------------------------------------------------
@@ -129,26 +129,32 @@ func (group *AbacoGroup) updateFrameTiming(p *packets.Packet, frameIdx FrameInde
 	if ts == nil {
 		return
 	}
+	newSubframeCount := frameIdx * abacoSubsampleDivisions
+	deltaSubframe := newSubframeCount - group.LastSubframeCount
 	deltaTs := ts.T - group.LastTimestamp.T
-	if frameIdx <= group.LastRowCount || deltaTs == 0 {
+	// There will be transient nonsense if the timestamp.Rate changes. I think the best
+	// approach is set subframe rate to 0. After the next (consistent) timestamp, it will recover.
+	unequalRates := ts.Rate != group.LastTimestamp.Rate
+	group.LastTimestamp = *ts
+	group.LastSubframeCount = newSubframeCount
+	if deltaSubframe <= 0 || deltaTs == 0 || unequalRates {
+		group.SubframeRate = 0
 		return
 	}
-	newRowCount := frameIdx * abacoSubsampleDivisions
-	deltaRow := newRowCount - group.LastRowCount
-	group.CountsPerRow = deltaTs / uint64(deltaRow)
-	group.LastTimestamp = *ts
-	group.LastRowCount = newRowCount
+	group.SubframeRate = deltaTs / uint64(deltaSubframe)
 }
 
-// rowCountFromTimestamp uses the group.FrameTimingCorrespondence data to
-// convert a timestamp in raw Abaco clock counts to a FrameIndex
-func (group *AbacoGroup) rowCountFromTimestamp(timestamp uint64) FrameIndex {
-	if group.CountsPerRow == 0 {
+// subframeCountFromTimestamp uses the group.FrameTimingCorrespondence data to
+// convert a timestamp (in raw Abaco clock counts) to a FrameIndex
+func (group *AbacoGroup) subframeCountFromTimestamp(timestamp uint64) FrameIndex {
+	if group.SubframeRate == 0 {
 		return 0
 	}
+	// This will be in error if the two timestamps have unequal Rates, but
+	// that problem should be rare and go away as soon as new Rate is reused.
 	deltaTs := int64(timestamp) - int64(group.LastTimestamp.T)
-	deltaF := deltaTs / int64(group.CountsPerRow)
-	return FrameIndex(int64(group.LastRowCount) + deltaF)
+	deltaF := deltaTs / int64(group.SubframeRate)
+	return FrameIndex(int64(group.LastSubframeCount) + deltaF)
 }
 
 func (group *AbacoGroup) enqueuePacket(p *packets.Packet, now time.Time) {
@@ -1138,7 +1144,7 @@ func (as *AbacoSource) extractExternalTriggers() []int64 {
 				// v := (u64data[i] >> 32) & 0xffffffff
 				// a := u64data[i] & 0xffffffff
 				t := u64data[i+1]
-				fc := int64(grp0.rowCountFromTimestamp(t))
+				fc := int64(grp0.subframeCountFromTimestamp(t))
 				// fmt.Printf("val 0x%8x  active 0x%8x  T 0x%16x    frameIndex %7d\n", v, a, t, fc)
 				externalTriggers = append(externalTriggers, fc)
 			}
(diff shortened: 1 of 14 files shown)


abaco.go
```diff
@@ -18,10 +18,10 @@ import (
 	"github.com/usnistgov/dastard/ringbuffer"
 )
 
-const abacoPseudoRowRate = 64 // The external triggers will be resolved this much finer than the frame rate
-// That is, we are writing external trigger info at a rate that we'd call the "row rate" in a TDM system.
-// So we write the data as if Abaco data sources had a 64-row readout, but all we mean by that is that
-// they are being resolved at 64x finer time steps than the frame rate.
+const abacoSubsampleDivisions = 64 // The external triggers will be resolved this much finer than the frame rate
```
ggggggggg, line 21: This should be `abacoSubframeDivisions`

abaco.go
```diff
@@ -18,10 +18,10 @@ import (
...
+// That is, we are writing external trigger info at a clock rate that we'd call the "row rate" in a TDM system.
```
ggggggggg, line 22: We are writing external trigger info at a clock rate finer than the frame rate, the "subframe" rate. 
In TDM systems the subframe rate is equal to the row rate, thus historical usage has conflated the two.

Proposed alternate comment.

abaco.go
ggggggggg: or `abacoSubframeRatio`

abaco.go
```diff
@@ -38,9 +38,9 @@ func gIndex(p *packets.Packet) GroupIndex {
 
 // FrameTimingCorrespondence tracks how we convert timestamps to FrameIndex
 type FrameTimingCorrepondence struct {
-	CountsPerRow  uint64
-	LastTimestamp packets.PacketTimestamp
-	LastRowCount  FrameIndex
+	SubframeRate      uint64 // Ratio of timestamp rate to subframe rate
```
ggggggggg, line 41: We should be consistent about `SubframeRate` vs `SubframeDivisions` vs `SubframeRatio`. Pick one and never use another. `SubframeTicks` is also reasonable.

abaco.go
```diff
@@ -129,26 +129,32 @@ func (group *AbacoGroup) updateFrameTiming(p *packets.Packet, frameIdx FrameInde
...
+	newSubframeCount := frameIdx * abacoSubsampleDivisions
+	deltaSubframe := newSubframeCount - group.LastSubframeCount
 	deltaTs := ts.T - group.LastTimestamp.T
-	if frameIdx <= group.LastRowCount || deltaTs == 0 {
+	// There will be transient nonsense if the timestamp.Rate changes. I think the best
+	// approach is set subframe rate to 0. After the next (consistent) timestamp, it will recover.
+	unequalRates := ts.Rate != group.LastTimestamp.Rate
+	group.LastTimestamp = *ts
+	group.LastSubframeCount = newSubframeCount
+	if deltaSubframe <= 0 || deltaTs == 0 || unequalRates {
```
ggggggggg, line 140: Wait are we allowing that the `SubframeRate` might change during acquisition? Or is this solely a learning step?

abaco.go
```diff
@@ -129,26 +129,32 @@ func (group *AbacoGroup) updateFrameTiming(p *packets.Packet, frameIdx FrameInde
...
+		group.SubframeRate = 0
 		return
 	}
-	newRowCount := frameIdx * abacoPseudoRowRate
-	deltaRow := newRowCount - group.LastRowCount
-	group.CountsPerRow = deltaTs / uint64(deltaRow)
-	group.LastTimestamp = *ts
-	group.LastRowCount = newRowCount
+	group.SubframeRate = deltaTs / uint64(deltaSubframe)
```
ggggggggg, line 144: What? Why would a computer timestamp be involved in calculating the `SubframeRate`?

ggggggggg: Or should `Timestamp` be called `Framestamp`?

data_source.go
```diff
@@ -583,7 +584,7 @@ func (ds *AnySource) HandleExternalTriggers(externalTriggerRowcounts []int64) er
 		ds.writingState.externalTriggerFileBufferedWriter = bufio.NewWriter(ds.writingState.externalTriggerFile)
 		// write header
 		msg := fmt.Sprintf("# external trigger rowcounts as int64 binary data follows, "+
-			"rowcounts = framecounts*nrow+row (nrow=%4d)\n", ds.rowFrameRatio)
+			"rowcounts = framecounts*nrow+row (nrow=%4d)\n", ds.subframeDivisions)
```
ggggggggg, line 587: subframecounts = framecounts*subframerate+subframeoffset

??

abaco.go
ggggggggg: I like `SubframeRate` the most after having gone most of the way through this review.

data_source_test.go
```diff
@@ -190,7 +190,8 @@ func TestWritingFiles(t *testing.T) {
 	if ds.processors[0].DataPublisher.HasLJH3() {
 		t.Error("Stop did not result in !HasLJH3")
 	}
-	if true { // prevent variable from persisting
+
+	if true { // block prevents variable from persisting (what variable?)
```
ggggggggg, line 194: maybe `err2`?

lancero_source.go
```diff
@@ -336,12 +336,16 @@ func (ls *LanceroSource) PrepareChannels() error {
...
+	ls.subframeOffsets = make([]int, ls.nchan)
 	index := 0
 	cnum := ls.firstRowChanNum
 	thisColFirstCnum := cnum - ls.chanSepColumns
 	ls.groupKeysSorted = make([]GroupIndex, 0)
 	for _, device := range ls.active {
-		ls.rowFrameRatio = device.nrows
+		// For Lancero sources, subframeDivisions = the number of rows.
+		// For sources with multiple LanceroDevice objects, its meaning is ambiguous, but we'll
+		// take the max value of all numbers of rows and hope for the best.
```
ggggggggg, line 347: Assert that they all have the same number of rows, throw an error if not. Then use that number of rows.

ggggggggg (commented):
I highly approve of this goal, and I think the number of comments I have about still remaining language ambiguity highlights the value of this PR. See the comments, there are still multiple different names for `SubframeRate` and `row` still appears in some places it shouldn't.

abaco.go
joefowler: Will do

abaco.go
joefowler: Good; will use

abaco.go
joefowler: Hmm, I was of the same mind at first. Then I decided that "rate" implies units of inverse seconds. This is a unitless quantity that divides the frame. Ratio is reasonable, but Divisions is what I settled on. I'll use it here.

abaco.go
joefowler: I see this as entirely a learning step. But if it did change, this would catch it fast. 

What I really had in mind, though, is not a change in _rate_, but the possibility that somehow the _alignment_ between counters gets off (maybe a dropped packet could do that?). This would allow the two to re-synchronize quickly.

abaco.go
joefowler: Well, Mazsi's label was "timestamp", but it's closer to a "firmware base clock counter". It counts in 4 ns increments (at least, the current one does).

data_source.go
joefowler: Yeah, why not?

data_source_test.go
joefowler: Well, I don't see it. I wonder if it was originally my comment? Nope, it was you, on October 18, 2018: https://github.com/usnistgov/dastard/blame/98232542e6d78fe689adb21def07cf8208e1011b/data_source_test.go#L193

joefowler: My guess is the comment is no longer accurate? I'll remove the block, assuming it still passes tests.

joefowler: I mean, I'll keep the _contents_ of the block, but remove the nesting into a pointless block.

lancero_source.go
joefowler: I cannot see ever exercising this case (would be reading 2+ lancero devices at once, with unequal # of rows). You convinced me that an assertion here is superior to what I was trying (a placeholder that would lead to great confusion if ever actually used). I'll assert/panic.

The **real** requirement here would be to treat external triggers as a per-device property, not a global property of the `LanceroSource`. But then we'd have to figure out how to write multiple external trigger files, which is a condition we _never_ anticipate using in the real world. Better to assert equality and panic if not present.

joefowler (commented):
Thank you hugely for the careful review, @ggggggggg !

abaco.go
ggggggggg: But doesn't that mean there was data corruption? I guess we've decided to allow dropped packets so maybe then? Can we at least print something out if it changes so there is some breadcrumb.

abaco.go
ggggggggg: Lets give it a name like `firmware_base_clock_counter` then. at least `firmware_timestamp` something to make very clear what it is, and that it is not `time.time()`

abaco.go
ggggggggg: I think `SubframeRate` makes sense because the rate has units inverse frames.

Issue #338, opened 2024-02-09
Dastard should write less crap to the terminal


joefowler:
Dastard is 7 years old. It doesn't need every darn debug line printed to the terminal. 

**Proposal**
* Stop writing all status messages to the terminal.
* Could write each broadcast message to `~/.dastard/logs/message_history.txt`
* Stop writing nearly all messages to the terminal.
* Maybe the terminal can have a line with curses-style info like <....*....> where the * bounces back and forth, one bounce with each heartbeat?

joefowler:
Fixed by #340.

PR #340, opened 2024-02-09
Nice terminal, without tons of meaningless/debug text

joefowler:
Instead, save the client messages to `~/.dastard/logs/updates.log` and put a cute "I'm still alive" ASCII bouncing text thing at the terminal. Only minimal terminal messages are printed now.


joefowler:
Fixes #338.

ggggggggg (approved):

PR #341, opened 2024-02-09
Accepting some comments from #339

joefowler:
I accidentally pulled PR #339 without noticing some suggestions. I agree that these ideas improve the variable names, so here is a revision.


ggggggggg (approved):
Unambiguous variable names!

PR #357, opened 2024-08-23
Don't allow default `Nsamples` or `Npresamples` of 0 to persist

### LINKED ISSUE #355: missing config leads to crash until you set the record length
This was dastard 0.3.4pre2 (git commit 4521992)

On the DR-TES flight computer, I can reproduce crashes with the following steps:

1) delete `.dastard` directory
2) start dastard
3) connect to ABACO (UDP) source (with 32 channels in this case but I don't know that matters)
4) turn on a auto trigger with delay 0 (for 1 channel)
5) after a short time, dastard crashes with error message below
6) restart dastard
7) connect to ABACO (UDP) source
8) after a short time, dastard crashes. Same error message, just different goroutine number

I then tried the same above steps, but with all 32 channels for the auto trigger. This time dastard starts to allocate huge amounts of memory after step 4.

If, after step 3, I set the record length to a non-zero value, I don't get the crash. Similarly, if I edit the `config.yaml` file to make `nsamples` nonzero, I don't get the crash.

I'm not using `dastardcommander` for these tests, I'm using my own script to command dastard, that I'm writig for dr-tes. I'll attach that script.

Specifically, I run it like:
```
drtes.py udp 191.168.14.100 # to connect to UDP
drtes.py start_noise 1 # to start auto trigger
```

Error message:
```
This is DASTARD version 0.3.4pre2 (git commit 4521992)
Logging problems       to /root/.dastard/logs/problems.log
Logging client updates to /root/.dastard/logs/updates.log

2024/08/23 13:49:37 Dastard config file: /root/.dastard/config.yaml
2024/08/23 13:49:52 New client connection established
2024/08/23 13:49:52 Starting data source named ABACOSOURCE
Sample rate for chan [4096-4127] 46875 /sec determined from  198 packets: Δt=0.050432 sec, Δserial=197, and 12.000 samp/packet
2024/08/23 13:50:52 New client connection established
2024/08/23 13:50:59 New client connection established
Panic! internal buffersChan full, len 100, capacity 100
panic: internal buffersChan full, len 100, capacity 100

goroutine 14 [running]:
github.com/usnistgov/dastard.(*AbacoSource).readerMainLoop(0xc0000e1188)
	/dastard/abaco.go:1071 +0xdf7
created by github.com/usnistgov/dastard.(*AbacoSource).StartRun in goroutine 11
	/dastard/abaco.go:951 +0xd6
```
[drcom.txt](https://github.com/user-attachments/files/16729950/drcom.txt)

### LINKED ISSUE #356: missing config leads to sim pulse problems
This was dastard 0.3.4pre2 (git commit 4521992).

This is using the DR-TES flight computer.

If the `status:nsamples` config parameter in `.dastard/config.yaml` is 0 when you try to start the simulated pulse data source, dastard will start to allocate huge amount of memory (multiple GB).

This parameter defaults to 0 when you start dastard with a missing `config.yaml` file, which is what happens when a new dastard installation is first started.

After manually editing `config.yaml` and changing  `status:nsamples` to 500, and restarting dastard, the memory issue went away.

`status:npresamp` is also 0 in the default config, and I also changed it to a non-zero number, so it's possible that this is the problematic parameter, not `status:nsamples`

Note that I'm not using `dcom` for this - I'm using a script that I'm writing for Dr-TES. The script is attached. I'd run it like this:

`drtes.py sim 4 4`
[drcom.txt](https://github.com/user-attachments/files/16729970/drcom.txt)

joefowler:
When reading the viper settings, it's possible that `Nsamples` and `Npresamples` are not found, and are set to the Golang default integer values of zero. Immediately fix non-positive values after reading.

Unfortunately, I cannot figure out a way to test this automatically, b/c the testing framework doesn't have direct access to the Dastard internals, but instead only talks to a running server. We'd have to connect to and parse the messages to the status-message port. Not worth the trouble, I judge.

Fixes #355. Fixes #356.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -1,9 +1,10 @@
 ## DASTARD Versions
 
-**0.3.4** May 2024-
+**0.3.4** August 23, 2024
 * Update all package dependencies (issue 347).
 * Send which types of files are active in the report about writing data (issue 343).
 * Rename bad Makefile variable `LDFLAGS` -> `GOLINKFLAGS` (issue 350).
+* When no config file exists, set `Nsamples` to a nonzero value, preventing some crashes (issues 355, 356).
 
 **0.3.3** May 28, 2024
 * Make publisher more responsive: don't block data processors waiting on disk flush.
--- a/global_config.go
+++ b/global_config.go
@@ -35,7 +35,7 @@ type BuildInfo struct {
 
 // Build is a global holding compile-time information about the build
 var Build = BuildInfo{
-	Version: "0.3.4pre2",
+	Version: "0.3.4",
 	Githash: "no git hash computed",
 	Date:    "no build date computed",
 }
--- a/rpc_server.go
+++ b/rpc_server.go
@@ -305,15 +305,18 @@ type SizeObject struct {
 // ConfigurePulseLengths is the RPC-callable service to change pulse record sizes.
 func (s *SourceControl) ConfigurePulseLengths(sizes SizeObject, reply *bool) error {
 	*reply = false // handle the case that sizes fails the validation tests and we return early
-	UpdateLogger.Printf("ConfigurePulseLengths: %d samples (%d pre)\n", sizes.Nsamp, sizes.Npre)
 	if !s.isSourceActive {
 		return fmt.Errorf("no source is active")
 	}
+	if sizes.Nsamp <= 0 || sizes.Npre <= 0 {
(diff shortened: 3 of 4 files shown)


joefowler:
@ggggggggg You are invited to review this, but we aren't going to merge until @danbek actually tests out that this fixes his problems. I'd rather have him test than me.

ggggggggg (approved):

PR #358, opened 2024-08-27
Bahama Enhanced with RPC Mode for Updated Data Assembly

cjtitus:
Bahama is a useful tool for testing, but it is limited by its command-line-only interface, and lack of customization with respect to data generation. It would be more useful if you could select:
* Multiple pulse amplitudes
* Pulse Width
* Sinusoid Frequency/Amplitude
* Sawtooth Frequency/Amplitude

Unfortunately, it would be unwieldy to add so many options as command-line flags. And testing that requires simulating both noise and pulses would necessitate shutting down Bahama and starting it back up.

I have drafted an enhancement for Bahama that adds an RPC interface, in a similar style to the Dastard RPC interface, that allows for configuration of the Pulse, Sinusoid, Sawtooth, and Noise parameters, and also allows for starting/stopping the data generation.

I have also modified the pulse generation function to be more configurable, with an easy-to-understand width parameter, and the possibility to give multiple amplitudes.

This will almost certainly undergo further evolution, as I use it to test an entirely synthetic data-generation/acquisition/analysis pipeline for NSLS-II. However, I wanted to throw it out here as a draft for comment.

Further ideas I might implement:
* Find a better way to stop/start data generation without crashing Dastard (or requiring Dastard to be stopped first)
* More customization for pulse generation (counts per second?)
* Some way to simulate line widths (may be going too far...)
* Possibility to upload a real pulse directly
* A way to preserve the run-from-command-line functionality (a --run flag?)

Comments welcome.

### DIFF
--- a/cmd/bahama/bahama.go
+++ b/cmd/bahama/bahama.go
@@ -14,6 +14,9 @@ import (
 
 	"github.com/usnistgov/dastard/packets"
 	"github.com/usnistgov/dastard/ringbuffer"
+	"log"
+	"net/rpc"
+	"net/rpc/jsonrpc"
 )
 
 const packetAlign = 8192 // Packets go into the ring buffer at this stride (bytes)
@@ -50,6 +53,27 @@ type BahamaControl struct {
 	noiselevel float64
 	samplerate float64
 	dropfrac   float64
+	host       string
+	cancel     chan os.Signal
+	Pulse      PulseParams
+	Sawtooth   SawtoothParams
+	Sinusoid   SinusoidParams
+}
+
+type PulseParams struct {
+    Amplitudes []float64
+    Width      int
+}
+
+type SawtoothParams struct {
+    Amplitude float64
+    Period    int
+}
+
+type SinusoidParams struct {
+    Amplitude float64
+    Frequency float64
+    Phase     float64
 }
 
 // Report prints the Bahama configuration to the terminal.
@@ -111,6 +135,20 @@ func (control *BahamaControl) Report() {
 	}
 	fmt.Printf("Data will be the sum of these source types: %s.\n", strings.Join(sources, "+"))
 	fmt.Println("Type Ctrl-C to stop generating Abaco-style data.")
+
+	if control.pulses {
+        fmt.Printf("Pulse amplitudes: %v\n", control.Pulse.Amplitudes)
+        fmt.Printf("Pulse width: %d\n", control.Pulse.Width)
+    }
+	if control.sawtooth {
+        fmt.Printf("Sawtooth amplitude: %.2f\n", control.Sawtooth.Amplitude)
+        fmt.Printf("Sawtooth period: %d\n", control.Sawtooth.Period)
+    }
+    if control.sinusoid {
+        fmt.Printf("Sinusoid amplitude: %.2f\n", control.Sinusoid.Amplitude)
+        fmt.Printf("Sinusoid frequency: %.2f\n", control.Sinusoid.Frequency)
+        fmt.Printf("Sinusoid phase: %.2f\n", control.Sinusoid.Phase)
+    }
 }
 
 // interleavePackets takes packets from the N channels `inchans` and puts them onto `outchan` in a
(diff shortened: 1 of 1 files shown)


joefowler:
Umm. Well, this is way way beyond what I ever envisioned using Bahama for. You say the old Bahama is "limited by its command-line interface" as if that were a bad thing.

I see your to-do item of somehow restoring the CLI features. Are you saying that this version is no longer useable without an RPC caller to configure it? If so, then you are essentially proposing that anyone using it would need to run FOUR programs:
1. Dastard
2. Dastard commander
3. Bahama
4. Bahama commander
I guess that's only 33% worse than the old way.

If you are proposing to leave this program unable to run stand-alone, then why isn't this just a separate program that stands in addition to Bamaha?

I'm glad that Bahama , to my surprise, is _almost_ useful for whatever you are developing. I mainly made it to be sure I could parse the (annoying) TLV packets on the Dastard end! @ggggggggg : thoughts?

cjtitus:
I will restore the CLI features, and I will try to make the defaults compatible with what was there previously, so that if you just call Bahama the same way as you used to, it does the same thing.

I am open to making this a totally separate program, but it is still 80% of the Bahama code, so having that much duplication in a second program seems like a bad thing to do, unless we split both Bahama and Bermuda into a separate package so that they can actually import common code, etc. But this is more complicated.

The main motivation isn't necessarily the RPC, it's the ability to do two things, which I could have emphasized more in the motivation:
* Create pulses of multiple amplitudes (i.e, to simulate our calibration sample)
* Switch between pulses and noise seamlessly, without restarting Bahama or Dastard

Instead of running four programs, I am actually running two:
* Dastard
* TES Server (which bridges between Bluesky and the TES -- we are also running it at BESSY)

TES Server talks to Dastard, and with some modifications, also now talks to Bahama. In this way, I can do clever simulation things. I can have a "set noise triggers" function in TES Server that first tells Bahama to just output noise. And then "set pulse triggers" has Bahama output a bunch of pulses with different amplitudes. In this way, I can test the beamline GUI (which Galen will be familiar with from BESSY).

As an aside, one of the interesting consequences of high-fidelity testing is that really can't go halfway. The GUI I'm building needs to display whether or not the TES is calibrated, and has projectors loaded. So to test that, I need to be able to take noise and pulses, make/load projectors, and then run a calibration scan that has actual lines. It's actually *easier* to just make Bahama do all of that at the lowest level, and then run Dastard and my GUI "for real" rather than trying to build some fake testing work-around mode into my GUI (which would be more complicated, and result in worse tests!).

cjtitus:
It's also worth mentioning that in building in the ability to re-generate data on-the-fly, it would only take 15 more minutes to add a "load data" function, as mentioned, that would let you upload real pulse data into the data buffer. I think this would be _**extremely**_ useful for testing triggers.

I know the Bahama pulse shape function is sort-of a "toy" model which does not really look like a TES pulse, but I've found that the settings for the "New Triggering" are way less robust than I expected, and I'd like to explore if this is due to some weird feature of the Bahama pulses, or if we're secretly dropping a certain number of pulses on the floor at NSLS-II...

ggggggggg:
I'm very supportive of having more end to end testing capability. In my mind Bahama is not the right place for this, I think simulated pulse source (or a new source altogether) makes a lot more sense because:
1. dastard already has a RPC interface
2. dastard already has a concept of generating fake data (simulated pulse source, trianglesource)
3. bahama is really intended to test the parsing of TLV packets (umux systems use these, future TDM systems might use these), so it significantly broadens the scope of bahama in a direction orthogonal to it's core reason to exist to add these features

That said, if you want to do it in bahama, it will work fine. I'm not sure how I feel about merging it in, but I guess as long as it doesn't break the CLI maybe it's fine?

How would you feel about doing this all in simulated pulse source instead?

cjtitus:
The things I do not like about the simulated pulse source are:
* It means I need to touch Dastard, which is significantly more scary. Introducing a subtle bug in Bahama is fine. Introducing one in Dastard is bad.
* It means that Dastard _itself_ is not really being tested, which will be a significant motivation later on.
* In the best case, I would like to be able to run Bahama in such a way that the _abaco_ settings in Dastard are identical to the beamline version, so that we have as few changes as possible between testing and production.

To flesh out the vision -- for a 10,000 pixel detector at NSLS-II, I believe we are probably going to have to run multiple copies of Dastard on a virtual machine cluster. Fibers will have to go from each read-out box to a heavy-duty network switch. We will need to make sure there are no bottlenecks anywhere in this process, from the networking to the VM cluster to the disk-writing.

I can emulate this end-to-end by running Bahama on a box, running fiber from that box to our network switch, putting Dastard on a VM cluster, etc. You lose a substantial chunk of the testing if we generate simulated pulses in Dastard.

This testing case is a bit in the future, but the future is really only about a year or two away, so best if I can get on top of things now, modify Bahama for end-to-end testing at the beamline scale, and then everything will be ready to start really stressing our IT systems _before_ we get 1000's of pixels.

cjtitus:
It's worth getting clarification on this -- it seems like from Dastard's point of view, getting pulses from "Bahama" is identical to getting them from the "real" hardware. Is this actually true? I have been assuming so.

ggggggggg:
> It's worth getting clarification on this -- it seems like from Dastard's point of view, getting pulses from "Bahama" is identical to getting them from the "real" hardware. Is this actually true? I have been assuming so.

Yes, this is as true as we can make it. 

Your case for putting this into bahama is pretty strong. I suggest you make `bermuda` which lives alongside `bahama` but doesn't have the backwards compatibility requirement for `bahama` CLI and then go full steam ahead and we'll start merging it into the dastard main branch.

PR #363, opened 2025-04-22
Fix zero ext trigger times

### LINKED ISSUE #362: HEATES reports all zeros in their external trigger timestamp file
@danbek and HEATES report that external triggers are being recorded to a binary file, as expected. But all the timestamp values are zero.

joefowler:
Fixes #362. Or, I think it does.

The correct behavior is to ignore packets that don't have an increased value of the subframe counter, when using them for `AbacoGroup.updateFrameTiming(...)`. We were erroneously setting certain values to zero when such (non-increasing) packets were found. Don't understand the firmware behavior enough to know what the cause was, but we do want to skip them for this purpose.

### DIFF
--- a/abaco.go
+++ b/abaco.go
@@ -134,9 +134,13 @@ func (group *AbacoGroup) updateFrameTiming(p *packets.Packet, frameIdx FrameInde
 	deltaSubframe := newSubframeCount - group.LastSubframeCount
 	deltaTs := ts.T - group.LastFirmwareTimestamp.T
 	// There will be transient nonsense if the timestamp.Rate changes. I think the best
-	// approach is to ignore. After the next (consistent) timestamp, it will recover.
+	// approach is set subframe rate to 0. After the next (consistent) timestamp, it will recover.
+	unequalRates := ts.Rate != group.LastFirmwareTimestamp.Rate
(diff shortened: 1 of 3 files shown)


abaco.go
```diff
@@ -133,15 +133,14 @@ func (group *AbacoGroup) updateFrameTiming(p *packets.Packet, frameIdx FrameInde
 	newSubframeCount := frameIdx * abacoSubframeDivisions
 	deltaSubframe := newSubframeCount - group.LastSubframeCount
 	deltaTs := ts.T - group.LastFirmwareTimestamp.T
+	if deltaSubframe <= 0 || deltaTs == 0 {
+		return
```
ggggggggg, line 137: I would suggest that we at least LOG the event, and maybe check that whatever values we care about other than the timestamp increase are the same between this packet and the last packet. If we're just getting 2 identical (for our purposes) packets, sure lets go ahead and ignore one. But if they disagree about what is happening, I am not comfortable going forward without better understanding. Eg what if one is right and one is wrong and our experimental results depend on picking the right one.

ggggggggg (changes requested):

ggggggggg (approved):

PR #369, opened 2025-05-11
Fix data panics when triggered records exceed array bounds

### LINKED ISSUE #360: Dastard crash when data dropped and trigger rate is high
Error when we have a very high photon rate at BESSY-II. Seen twice:
```text
DATA DROP. firstFrameIndex 0, droppedFrames 248
panic: runtime error: slice bounds out of range [-1:]

goroutine 28700547 [running]:
...
```

The error trace says error is in `triggering.go:48`, which reads 
```go
copy(data, stream.rawData[i-NPresamples:i+NSamples-NPresamples])
```
Clearly we are trying to go back before the existing data, and apparently by exactly 1 sample. Sounds like an off-by-1 error somewhere?? See if we can debug on a live machine, but also try to sniff it out.

joefowler:
Fixes #360 or works around it. I'm not 100% sure. But we needed this change to operate successfully at BESSY. I'm adding it to the main line of code now.


ggggggggg (commented):
Not super elegant, but looks like it work around a real annoying problem so approve.

PR #370, opened 2025-06-10
Reorganize code, moving internal packages to recommended `internal` directory

joefowler:
The recommended way to lay out a go module that has features organized into sub-packages is to put them in `internal/`, whenever their API should not be available outside the current package. See [Organizing a go module](https://go.dev/doc/modules/layout). Unless there's a specific need to share its API (and the maintainers are willing to keep it up), sub-packages should be in `internal/`.

So I have moved all our sub-packages with the sole exception of `packets`. This handles the UDP data packets generated by the µMUX FPGA. I can imagine using this outside of Dastard, so I'll leave it out of the internal directory.

Other than moving files and changing the import specs, nothing should be different here.


joefowler:
> I don't really see much benefit here, but it's also following some guidance and doesn't seem much worse, so I'll approve.

I don't think there is benefit, but it seemed like a good time to follow Go guidance.

PR #371, opened 2025-06-17
Synchronize files in a better way

joefowler:
Not ready to merge this yet, but we should try this way of keeping files synced up. Was discussed last fall, but never pulled into main Dastard.


ggggggggg:
> Also, it does nothing with OFF files. Is that okay with everyone?

No I don't think that is ok. We should do it with each type of file we can write.

ggggggggg (approved):

PR #379, opened 2025-12-05
Implement better non-blocking writes

joefowler:
This is ready to try in real life, I think.

I uses a non-synchronizing, infinite-size, channel-like queue called `unboundedChannel` to store slices of pointers to data records, for later writing in LJH 2.2, 3.0, or OFF formats. The behavior allows the publisher object to "fire and forget" the records.

### DIFF
--- a/process_data.go
+++ b/process_data.go
@@ -84,13 +84,12 @@ func NewDataStreamProcessor(channelIndex int, broker *TriggerBroker, NPresamples
 		stream: *stream, 
 		NSamples: NSamples, 
 		NPresamples: NPresamples,
-		DataPublisher: *NewDataPublisher(),
 	}
-	fmt.Printf("New DSP: %p  with publisher %p\n", &dsp, &(dsp.DataPublisher))
 	dsp.LastTrigger = math.MinInt64 / 4 // far in the past, but not so far we can't subtract from it
 	dsp.projectors = &mat.Dense{}       // dsp.projectors is set to zero value
 	dsp.basis = &mat.Dense{}            // dsp.basis is set to zero value
 	dsp.EMTState.reset()                // set up edgeMulti in known state
+	dsp.DataPublisher.StartWriteDataLoop()
 	return &dsp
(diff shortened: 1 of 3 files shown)


internal/unboundedchan/unboundedchan.go
```diff
@@ -0,0 +1,82 @@
+package unboundedchan
```
ggggggggg: I'm very against adding unbounded chan, I think it solves no problem and will make debugging harder. What happens when we run out of memory because an unbounded chan is so full of crap that everything stalls? How do we know what happened?
So then we should add some monitoring so we get a warning if the unbounded chan is too full? Well why not use a bounded chan at that point. 

It's also more code we need to maintain that may contain bugs.

ggggggggg (commented):
It seems like the point of this is to let us write in a "fire and forget" mode where writes can't delay progress in the rest of the code. The existing and deleted by this PR asyncbufio also had the same goal, did it work or not work? Why is this better than it? I'm not a huge fan of asyncbufio, but if we're going to churn to replace it we should have some concrete analysis of how this is better. It just seems different. Maybe asyncbufio is fundamentally flawed and should be removed. 

I am strongly against adding unbounded channels, I think they don't have any practical value over bounded channels with large bounds, and have many potential downsides. I think the only benefit of an unbounded channel over a bounded channel with large bound is minimum memory usage. You pay for the space for all the pointers you would ever hold onto all the time with a bounded channel, and that is basically free for us.

I am strongly against adding a mutex to data publisher, as it goes backwards from our progress towards "structured concurrency" where we avoid race conditions by having work start and stop at defined points and reducing the number of long running go routines.

Finally, this approach of buffering will only ever help us with temporary throughput problems, but not maximum throughput problems, unless it helps us make larger write calls. At least that was my conclusion from using fio to benchmark writing to disk. Lots of small write calls to different files was always slow on a spinning disk, but the throughput was good if we made fewer large write calls.

PR #380, opened 2025-12-05
Verify UDP buffer minimum size when running with Abaco UDP sources

joefowler:
This check is only made when:
* On a Linux computer
* You start an AbacoSource
* You are using at least UDP AbacoDevice (ring-buffer devices don't count).

Goal is to force you to check the README about increasing the receive buffer size.


ggggggggg (changes requested):
Looks good, I just think we should give the action item in the panic message.

ggggggggg (approved):

PR #390, opened 2026-02-02
be slightly more forgiving on sample rate difference

ggggggggg:
I kept getting sample rate errors using a single RFSoC on myriad, so I just made it complain less. Not really sure why this happens, but it will still catch big errors.


joefowler (approved):
Hmm. Is it possible that your problem is nothing more than roundoff error, or similar? For instance, two F-sets generate different packet sizes, and equal sample rates generate floating-point numbers that differ by a meaningless (but nonzero) amount? If so, then loosening this requirement sounds perfectly reasonable.

PR #391, opened 2026-02-24
Bahama hostname

cjtitus:
# Update to command line options for Bahama

When running on distributed systems, it may be necessary to bind bahama to hosts other than `localhost`.

I have just added a host flag to the argument parser, and then threaded the host argument through. If no host is given, the default hostname is "localhost", as before.

This has been tested in docker containers, and works fine.

### DIFF
--- a/cmd/bahama/bahama.go
+++ b/cmd/bahama/bahama.go
@@ -299,7 +299,7 @@ func main() {
 	changaps := flag.Int("gaps", 0, "How many channel numbers to skip between groups (relevant only if ngroups>1 or nsource>1)")
 	ngroups := flag.Int("ngroups", 1, "Number of channel groups per source, (1-Nchan/4) allowed")
 	port := flag.Int("port", 4000, "UDP port to produce data")
-	host := flag.String("host", "0.0.0.0", "Hostname or IP address to bind to")
+	host := flag.String("host", "localhost", "Hostname or IP address to bind to")
 	samplerate := flag.Float64("rate", 200000., "Samples per channel per second, 1000-500000")
 	noiselevel := flag.Float64("noise", 0.0, "White noise level (<=0 means no noise)")
 	usesawtooth := flag.Bool("saw", false, "Whether to add a sawtooth pattern")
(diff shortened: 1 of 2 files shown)


cmd/bahama/bahama.go
```diff
@@ -355,7 +357,7 @@ func generateAndPublishData(control BahamaControl, cancel chan os.Signal) {
 		packetchan := make(chan []byte)
 		defer close(packetchan)
 		portnum := (control.port) + cardnum
-		if err := udpwriter(portnum, packetchan); err != nil {
+		if err := udpwriter(control.host, portnum, packetchan); err != nil {
 			fmt.Printf("udpwriter(%d,...) failed: %v\n", portnum, err)
```
joefowler, line 361: I suggest changing to 
```go
fmt.Printf("udpwriter(%s,%d,...) failed: %v\n", control.host, portnum, err)
```

That is, add the hostname to the arguments printed in the error message.

joefowler (changes requested):
This looks good to me, though I'm worried about the error in the automated testing. Do you understand that?

Aha! You need to change line 129 in `cmd/bahama/bahama_test.go` to take the new arguments. Ideally, also change the following line with its error message. See
https://github.com/cjtitus/dastard/blob/43bf932bb4bf0e0dc835ba6357f9c0566f0b10cb/cmd/bahama/bahama_test.go#L129-L130

ggggggggg (approved):

joefowler:
> This looks good to me, though I'm worried about the error in the automated testing. Do you understand that?
> 
> Aha! You need to change line 129 in `cmd/bahama/bahama_test.go` to take the new arguments. Ideally, also change the following line with its error message. See https://github.com/cjtitus/dastard/blob/43bf932bb4bf0e0dc835ba6357f9c0566f0b10cb/cmd/bahama/bahama_test.go#L129-L130

@cjtitus : I'm not going to make this change, because you have done your work in a fork of your own. But when you have a chance to fix the failing tests, I'm happy to merge this.

cjtitus:
I have updated the tests. Let me know if you want me to re-base first. I can't imagine that my changes conflict with anything though.

joefowler (approved):

Issue #403, opened 2026-08-17
Generalize the "Abaco" data source to work with TKID and KICS resonator readouts


joefowler:
The existing "Abaco" source should be easy to generalize so that it works with other resonator readouts: TKIDs and KICS systems. It already works with non-Abaco RFSoC systems. Thus, the old name is already a misleading historical legacy, about to get worse. What all its data systems have in common:

- We’re reading out resonators
- The quantity of interest is an angle in some abstract space
- Data arrive in the form of UDP packets
- …designed to the Imrek TLV spec
- µMUX requires a software correction for potentially lost flux quanta; KICS and TKID don't.

We'll want to rename "abaco" throughout, possibly as a "resonator readout", though I'm still open to ideas.

PR #404, opened 2026-08-18
"Resonators" instead of "Abaco"

joefowler:
- Rename objects and files named Abaco to "Resonator". 
- Make the default behavior of the `PhaseUnwrapper` be to work even if there are no unwrapping required, so that it can still be there to invert pulses. (In spite of the name, the `PhaseUnwrapper` also inverts pulses.)
- Make it okay to have a nil unwrapper. If nil, it's simply skipped.

This version of Dastard will be tagged as v0.3.10, and it will require updates to D-com, too.


joefowler:
Fixes #403

